PR #1, opened 2017-08-31
[WIP] Move identity into a separate area

javiercn:
I will not be merging this PR, but using the changes here to showcase what we need to do once we have a 2.1 templates branch/folder.

@SteveSandersonMS https://github.com/aspnet/templating/commit/2d7fa642b7808503949d1eb8faaebcf8875fa4dd is something that we might actually want to do.


SteveSandersonMS:
@javiercn I'm not sure how useful 2d7fa642b7808503949d1eb8faaebcf8875fa4dd is, since we'll already have nightly builds of the template package corresponding to the nightly build of ASP.NET Core itself. Also (to support that) the tokens like `$(TemplateAspNetCoreVersion)` will be evaluated during the build process so they won't still be present post-build, so it doesn't look like this implementation would work, unless I'm misunderstanding your use case.

If this PR is not intended to be merged, could you close it when you're done with it?

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -111,11 +119,35 @@
         {
           "choice": "Windows",
           "description": "Windows authentication"
+        },
+        {
+          "choice": "IndividualTokenService",
+          "description": "Individual authentication with a token server"
```
Eilon: Do these strings show up to users? If so, need to get PM sign-off

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
```diff
@@ -111,11 +119,35 @@
...
         }
       ],
       "defaultValue": "None",
       "description": "The type of authentication to use"
     },
+    "AspNetCoreVersion": {
+      "type": "parameter",
+      "datatype": "string",
+      "description": "The version for AspNet Core packages",
```
Eilon: Does this show up to users?

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
```diff
@@ -0,0 +1,465 @@
...
+        [HttpGet]
+        [AllowAnonymous]
+        public async Task<IActionResult> LoginWith2fa(bool rememberMe, string returnUrl = null)
+        {
+            // Ensure the user has gone through the username & password screen first
+            var user = await _signInManager.GetTwoFactorAuthenticationUserAsync();
+
+            if (user == null)
+            {
+                throw new ApplicationException($"Unable to load two-factor authentication user.");
```
Eilon: We need to get rid of all of these bogus ApplicationExceptions. MSDN has said not to use them since like 2002, but I think somehow they snuck into the templates a while ago 😢

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/NuGet.config
```diff
@@ -0,0 +1,8 @@
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="AspNetCore" value="https://dotnet.myget.org/F/aspnetcore-ci-dev/api/v3/index.json" />
+    <add key="AspNetCoreTools" value="https://dotnet.myget.org/F/aspnetcore-tools/api/v3/index.json" />
```
Eilon: Why is this feed needed here?

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
javiercn: Nope. They are hidden parameters on the CLI so they don't show in dotnet new.

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
javiercn: I'm collecting all these small pieces of feedback and will file a bug with all the things to fix on the template.

javiercn:
@SteveSandersonMS 2d7fa64 is for being able to instantiate templates against whats currently on the dev feed and being able to work without having to build a nupkg for the templates. I was running into a lot of trouble iterating over the template changes and this makes it a lot easier. Maybe (probably) i'm doing something wrong and there might be an easier way to do this. I'll talk to you about this next time I see you.

HaoK:
@javiercn you aren't able to just run dotnet new against the feed?  `dotnet new -i C:\GitHub\templating\template_feed`

That's all I did in the past to test local changes

javiercn:
@HaoK Nope. You need to build so that the versions get replaced afaik and a .nupkg is built. That's why I want to have these settings so that you can just do something like dotnet new -i .\templating\src\Microsoft.DotNet.Web.ProjectTemplates.<<version>>\content; ```dotnet new mvc -au Individual -acv 2.1.0-* -efcv 2.1.0-*```

If you look at https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Company.WebApplication1.csproj#L31 the versions are replaced during build and a nupkg with fixed versions is created (the only way to create such nupkg I've found is through the command line build.cmd which does a full build and has to pass all the tests for the artifacts to get produced, which slows the process of iterating over the template). I want to avoid having to do that by just being able to point to the content of the template and get the packages from our dev feed. I don't know if there is a better/faster way of doing this, but I think both ways are complementary and this doesn't cause any issue as it's completely hidden for our customers. (None of these parameters show up on any UI) and as I said, it makes our lives way easier (unless there's a trick that I don't know)

src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/.template.config/template.json
Eilon: Ok. Should still fix to the proper product name `ASP.NET Core`.

HaoK:
Sounds like we just want the dev template versions to be 2.1.* instead of a fixed version to pickup the latest?

SteveSandersonMS:
> [Javier] the only way to create such nupkg I've found is through the command line build.cmd which does a full build and has to pass all the tests for the artifacts to get produced, which slows the process of iterating over the template

You don't have to go through the whole `build.cmd` process to get new template nupkgs locally. You can either:

 * Run `dotnet pack` at the repo root to rebuild all the template nupkgs. This takes 10-15 seconds.
 * Or, run `dotnet pack <project>` to rebuild a specific template nupkg. For example, `dotnet pack src\Microsoft.DotNet.Web.ProjectTemplates.2.0`. This completes in about 2 seconds.

Is that sufficient for your needs?

javiercn:
Closing this as another PR was already merged on the correct branch/version

PR #13, opened 2017-09-21
Remove IncludeScopes from logger configurations

pakrym:
Merge filter rules too.


glennc:
False is the default right?

pakrym:
Yes

glennc:
@mlorbetske Is this still the right place to do this? I heard talk of moving it.

PR #16, opened 2017-09-22
Fix Aurelia template to lower the entry level

MaximBalaganskiy:
Added jQuery and made it available globally
Added Promise polyfill to support IE ootb
Fixed vendor bundle to have correct format modules

### DIFF
--- a/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/webpack.config.vendor.js
+++ b/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/webpack.config.vendor.js
@@ -2,6 +2,7 @@ var path = require('path');
 var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var extractCSS = new ExtractTextPlugin('vendor.css');
+const { AureliaPlugin, ModuleDependenciesPlugin } = require("aurelia-webpack-plugin");
 
 module.exports = ({ prod } = {}) => {
     const isDevBuild = !prod;
@@ -44,6 +45,7 @@ module.exports = ({ prod } = {}) => {
         },
         plugins: [
             extractCSS,
+            new AureliaPlugin({ aureliaApp: undefined, dist: "native-modules" }),
             new webpack.ProvidePlugin({ $: 'jquery', jQuery: 'jquery' }), // Maps these identifiers to the jQuery package (because Bootstrap expects it to be a global variable)
             new webpack.DllPlugin({
                 path: path.join(__dirname, 'wwwroot', 'dist', '[name]-manifest.json'),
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
Thanks for the contributions!

For us to be able to include some of these changes, we'd need more information on what benefits they offer (or possible drawbacks). Ideally, it's best to first open an issue at https://github.com/aspnet/javascriptservices/issues (which is where we design and plan the SPA templates - I know this isn't obvious though!) to discuss the proposals.

As for this PR, we probably wouldn't want to expose `jQuery` to global scope as that's not great practice in modern SPA development. For the IE shims, this is something we'll be addressing in a common way across all the SPA templates, so you don't need to cover that. I'd like to know what `AureliaPlugin` does and why it was necessary to import `ModuleDependenciesPlugin` even though it doesn't seem to be used. And finally, I know this is a minor detail, but the code style doesn't quite match the rest of the file (e.g., use of `"` quote and `const`/`var`).

I'll close this pending your further discussion at https://github.com/aspnet/javascriptservices/issues. Hope that's OK!

PeterWone:
The don't-strew-your-junk-around-the-global-scope rule exists to prevent accidental collisions. Just about everything ever written depends jQuery. Where is it expected to be? **In the global scope, as jQuery and as $.** If you won't let something as ubiquitous as jQuery use it, the global scope might as well not exist.

SteveSandersonMS:
@PeterWone We're not stopping people from putting jQuery in global scope if they want. That's a choice for each developer. It's just not required by default.

PeterWone:
@SteveSandersonMS  I understand your position quite well and for me or Max it's fine. The flaw with this approach is that it only caters to the needs of people who understand the issues we're discussing. I don't believe these are the primary users of your templates. I could be wrong about that; I have no statistics, perhaps you do.

src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/webpack.config.vendor.js
```diff
@@ -2,6 +2,7 @@ var path = require('path');
 var webpack = require('webpack');
 var ExtractTextPlugin = require('extract-text-webpack-plugin');
 var extractCSS = new ExtractTextPlugin('vendor.css');
+const { AureliaPlugin, ModuleDependenciesPlugin } = require("aurelia-webpack-plugin");
```
AdamWillden: `ModuleDependenciesPlugin` is unused

MaximBalaganskiy: yeah, that's a leftover... should be removed. Much of the PR is not getting in, does it make sense to just create a new one?

SteveSandersonMS:
> does it make sense to just create a new one?

@MaximBalaganskiy Yes, if that's OK. Thanks!

PR #38, opened 2017-10-01
feat(package): updated all libraries

claylaut:


Eilon:
@SteveSandersonMS please review & merge if this is ok.

SteveSandersonMS:
Thanks for the suggestion!

As it happens, we're overhauling the Angular template more broadly by moving to an Angular CLI-based template. In this process we'll be getting a new `package.json` file that references whatever is the latest near the time we're ready to ship.

So if it's OK with you @claylaut I'll close this now as the equivalent change will happen automatically during our update.

claylaut:
@SteveSandersonMS do you have an approximate idea when this will be merged/released? Will it be in angular 5?

PR #41, opened 2017-10-02
chore(scripts): add useful npm scripts shorthands

stephenlautier:
This will help development process when working with this template.

You can build/launch app via:

```bash
# build all as dev
npm run build 

# build all as rel
npm run build:rel

# launch (this is even more useful when having launch profiles)
npm start
```


Eilon:
This seems reasonable. Would it make sense to apply the same change to the other SPA templates?

stephenlautier:
Yes it makes more sense if they are consistent, I will do the changes for the rest

stephenlautier:
@Eilon changes for all other SPA templates pushed.

I would suggest adding a `README.md` with the template and perhaps add some useful commands in order to be easier getting started (even something simple like my first comment) and perhaps some link to https://github.com/aspnet/JavaScriptServices as well. If you agree I can do that as well (or maybe I can do it as another PR).

Eilon:
@stephenlautier thanks! Yeah if you can send another PR for the README changes that would be great.

@SteveSandersonMS can you review & merge?

SteveSandersonMS:
Thanks for the suggestions @stephenlautier!

We're currently in the process of changing how the build steps are arranged. This is mainly due to moving to an Angular CLI basis for the Angular template, and after that trying to make the others more consistent with that new arrangement. We are in fact [using NPM scripts already](https://github.com/aspnet/templating/blob/413c3deb72daea3f3abc210f7e768786fd5843e0/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package.json#L8) in that new arrangement, though the exact final details are still emerging.

We'll need to review this PR at some later date when we can determine to what extent it applies.

SteveSandersonMS:
As per the preceding comment, we're handling this as part of the broader template changes. For example, https://github.com/aspnet/templating/blob/feature/stevesa/angular-cli-template/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package.json#L7

Since this is being covered, and since the diffs in this PR won't apply to the newer project structures, I'll mark this as closed. Hope that's OK!

stephenlautier:
It's not a problem, as long as we are going to have something like this anyway 👍 . 
Looking forward to the new changes!

PR #46, opened 2017-10-05
Vue.js template: Use async/await

muratg:


muratg:
@SteveSandersonMS FYI

Eilon:
@muratg is this the only place in the app it needs to be updated?

muratg:
@Eilon I took a quick look at the other SPA templates and this one was the only one I could find.  I'll take another look and will also take a look at other templates as well.

SteveSandersonMS:
Thanks @muratg! This is now merged.

I made a couple of [minor tweaks](https://github.com/aspnet/templating/commit/5d7e31d16f013c05f5b8b45d49dafa63f3c4f290) - hope that's OK. Removed the `try/catch` because this can be handled upstream, and replaced a `let` with `const` for clarity.

Issue #31, opened 2017-09-22
Templates should use app.UseMvcWithDefaultRoute();


mlorbetske:
_From @mlorbetske on August 3, 2017 17:56_

_From @Rick-Anderson on July 28, 2017 2:6_

I recommend the templates use `app.UseMvcWithDefaultRoute();` rather than

```csharp
app.UseMvc(routes =>
{
    routes.MapRoute(
        name: "default",
        template: "{controller}/{action=Index}/{id?}");
});
```

The more complex code is a barrier to entry for new users who don't need this flexibility. Advanced users will have no problem adding the above code when they need it. The code above doesn't make much sense to Razor Pages developers. Recommend this change for all templates but especially for the Razor Pages template.

cc @DamianEdwards  - part of the lower the barrier to entry effort.

_Copied from original issue: aspnet/Templates#857_

_Copied from original issue: dotnet/templating#1180_

mlorbetske:
_From @Rick-Anderson on August 8, 2017 19:2_

The Razor Pages template should use `app.UseMvc();`

Why add controller routing to a RP project.  @rynowak  - do you agree?

Rick-Anderson:
@rynowak  can you comment on RP routing in templates?

rynowak:
> The Razor Pages template should use app.UseMvc();

Yes

guardrex:
@scottaddie In addition to your #47, will you (or someone) also put one in for `UseMvcWithDefaultRoute` for the MVC template? ... It has a bearing on what they're doing with the Azure-Docs samples. I'm recommending it to them now; so if the decision is made to go that way for 2.1, they'll probably take that as an approval to go forward with `UseMvcWithDefaultRoute` today. It'll be one less thing to update later if they do.

Rick-Anderson:
The issue title is 
**Templates should use `app.UseMvcWithDefaultRoute();`**
- with the exception of RP templates. So yes, the other templates should be changed.

danroth27:
@Eilon 

The reason the default route is explicitly configured is because we assume that for MVC apps you will want to add and modify routes, so we want the template to give you an obvious place to do that. I think we should leave the route as is for the MVC template.

The reason the same route is present in the Web API and Razor Pages repos is because you might decide to add or scaffold MVC controllers into your app. I think this reason is less compelling, and I would be ok switching to `UseMvc()` (no convention based routes) for these templates.

Rick-Anderson:
[...]

PR #47, opened 2017-10-06
Invoke a simpler UseMvc overload in Razor Pages templates

scottaddie:
Fixes https://github.com/aspnet/templating/issues/31.

My assumption is that we only want to apply this change to the Razor Pages project templates. It's Razor Pages, after all, that's intended to lower the barrier to entry.

/cc: @Rick-Anderson @DamianEdwards @rynowak


Eilon:
This seems like a good change. However, there's more to this in https://github.com/aspnet/templating/issues/31, so we should merge this PR, and then get to the rest of https://github.com/aspnet/templating/issues/31.

SteveSandersonMS:
Merged into both the `dev` and `release/2.0.0` branches (we changed how the repo is structured since this PR was submitted - the different versions of the templates are in different branches now, instead of being in version-numbered folders).

PR #57, opened 2017-10-11
Remove the 2.1. templates from the rel/2.0.x branch

natemcmaster:
The templates will stay in the dev branch. 

FYI - I'll do a non-merging merge after this goes in so we can keep rel/2.0.x work on the 2.0 templates in sync with the dev branch.


natemcmaster:
It's not technically necessary. It's more about keeping branches clean. I figured it would be better to keep the rel/2.0.x branch free of any netcoreapp2.1 stuff.

PR #58, opened 2017-10-11
Automate dependency-version selection

natemcmaster:
This uses the 2.0.x build tools to do what PackageLineup does in dev. (btw, this is also the way we'll do things in dev once we finish the 2.0.x patch).

Changes:
 - Updated to KoreBuild to 2.0.2-beta-15522
 - Renamed content/\*/\*.csproj files that have PackageReference in them to \*.csproj.in 
 - Replaced any hard-coded versions with MSBuild properties
 - Added targets to call GenerateFileFromTemplate on \*.csproj.in to generate content\*\*.csproj
 - Set TreatWarningsAsErrors to true in the projects generated during test to ensure default templates don't have build or restore warnings


natemcmaster:
@SteveSandersonMS the tests fail for me, something to do with selenium. The test is able to make a new project, build, and launch the server, but fails attempting to launch a browser. What's required to make these pass?

```
Failed   Templates.Test.SpaTemplateTest.SpaTemplate_Works(targetFrameworkOverride: null, template: "react")
Error Message:
 OpenQA.Selenium.WebDriverException : Cannot start the driver service on http://localhost:49710/
Stack Trace:
   at OpenQA.Selenium.DriverService.Start()
   at OpenQA.Selenium.Remote.DriverServiceCommandExecutor.Execute(Command commandToExecute)
   at OpenQA.Selenium.Remote.RemoteWebDriver.Execute(String driverCommandToExecute, Dictionary`2 parameters)
   at OpenQA.Selenium.Remote.RemoteWebDriver.StartSession(ICapabilities desiredCapabilities)
   at OpenQA.Selenium.Remote.RemoteWebDriver..ctor(ICommandExecutor commandExecutor, ICapabilities desiredCapabilities)
   at Templates.Test.Helpers.WebDriverFactory.CreateWebDriver() in C:\dev\aspnet\templating\test\Templates.Test\Helpers\WebDriverFactory.cs:line 23
   at Templates.Test.Helpers.AspNetProcess.VisitInBrowser() in C:\dev\aspnet\templating\test\Templates.Test\Helpers\AspNetProcess.cs:line 102
   at Templates.Test.SpaTemplateTest.TestApplication(String targetFrameworkOverride, Boolean publish) in C:\dev\aspnet\templating\test\Templates.Test\SpaTemplateTest.cs:line 40
   at Templates.Test.SpaTemplateTest.SpaTemplate_Works(String targetFrameworkOverride, String template) in C:\dev\aspnet\templating\test\Templates.Test\SpaTemplateTest.cs:line 28
```

build/dependencies.props
```diff
@@ -0,0 +1,46 @@
+﻿<Project>
```
SteveSandersonMS: I don't know how these numbers are determined or how they get updated, but I trust that's automated somehow as part of your replacement for lineups.

build/repo.targets
```diff
@@ -1,17 +1,8 @@
 <Project>
-  <Import Project="FixPackageOutputDirs.targets" />
-
   <UsingTask
     TaskName="PullRequestSubmitter.PullRequestTask"
     AssemblyFile="$(MSBuildThisFileDirectory)..\tools\PullRequestSubmitter\bin\Release\netcoreapp2.0\PullRequestSubmitter.dll" />
 
-  <Target Name="PushNuGetPackages">
```
SteveSandersonMS: Are you removing this only because it's not needed in the patch branch? This wouldn't be removed when you make the same changes to `dev`, would it?

korebuild-lock.txt
```diff
@@ -0,0 +1,2 @@
+version:2.0.2-beta-15522
+commithash:f26cb086ebd8b5a187381bb9a431794f11d69025
```
SteveSandersonMS: Going to need some tutorials on these changes to Korebuild when these updates are applied to all the `dev` branches :)

src/Microsoft.AspNetCore.SpaTemplates/.gitignore
```diff
@@ -0,0 +1,2 @@
+content/*/*.csproj
+content/*/*.fsproj
```
SteveSandersonMS: This system is a bit awkward because it looks like it prevents you from having a `.sln` that includes the template `.csproj`es. You'd need to be careful to run the `build` task to generate the `csproj` files before you could open the `.sln`.

Are you still considering whether you could make the original `csproj` files runnable directly like they were before?

build/dependencies.props
natemcmaster: There are two scenarios to this. 

1. Automatically flowing new versions when building aspnet/Templating in Universe. Universe generates a file with new versions and sets DotNetPackageVersionPropsPath. This will override versions set in the repo source code.

https://github.com/aspnet/templating/blob/c240aaa1aef89a3b5a6f765a9bb2229f10b1cbec/build/dependencies.props#L45

2. Updating the repo source code. We haven't automated this yet, but we plan to. Before we make these changes in dev, we'll make some command-line tools to make updating to latest easier. We'll also add some check to ensure the right variable names are used, floating package versions are not, and that that all PackageReference's use a variable.

test/Templates.Test/Helpers/TemplatePackageInstaller.cs
```diff
@@ -72,13 +73,13 @@ private static void ReinstallTemplatePackages(ITestOutputHelper output)
         private static void VerifyCannotFindTemplate(ITestOutputHelper output, string templateName)
         {
             // Verify we really did remove the previous templates
-            var tempDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString("D"));
+            var tempDir = Path.Combine(AppContext.BaseDirectory, Path.GetRandomFileName(), Guid.NewGuid().ToString("D"));
```
SteveSandersonMS: Out of interest, why do you want the `tempDir` to be inside `AppContext.BaseDirectory`?

natemcmaster: global.json. On my machine, I have a bunch of 2.1.0 stuff installed. When running without a global.json file, the tests pick up the latest installed dotnet-sdk. It caused failures because `dotnet new` without `--framework-override` would default to the netcoreapp2.1 templates.

SteveSandersonMS:
> the tests fail for me, something to do with selenium

It tries to use Edge if you're on Windows 10, or Firefox if you aren't. Is it possible your system doesn't have the necessary browser?

test/Templates.Test/Helpers/TemplatePackageInstaller.cs
SteveSandersonMS: Makes sense. Thanks.

src/Microsoft.AspNetCore.SpaTemplates/.gitignore
natemcmaster: I played around with options, but didn't find anything that worked well. The difficulty was in differentiating between MSBuild variables and template variables.  My primary concern was to remove the system in which we have to edit files after they are placed in the nupkg. If you have other ideas for a way to keep you existing workflow, let me know.

korebuild-lock.txt
natemcmaster: Good to know. It'll be a few weeks before these changes make it to `dev`. I'll send instructions to the whole runtime team before these changes happen.

natemcmaster:
I'm on the Windows 10 Fall Creators update. I'm guessing this means something in edge changed...

build/repo.targets
natemcmaster: Yeah, not needed in the patch branch. aspnet/Universe will push the packages to our patch myget feed.

SteveSandersonMS:
> I'm guessing this means something in edge changed...

Sounds very likely.

src/Microsoft.AspNetCore.SpaTemplates/.gitignore
SteveSandersonMS: What's so problematic about doing the substitution after the nupkg is created?

natemcmaster: > it looks like it prevents you from having a .sln that includes the template .csprojes

We could still have a .sln file for this. You would just have to execute `dotnet build` on `Templating.sln` first to generate the csproj files.

SteveSandersonMS: OK, understood. At the same time, this approach breaks other basic scenarios like opening a solution in VS (without first doing a special action). A key idea of the `dotnet new` project system is [runnable projects](https://github.com/dotnet/templating/wiki/%22Runnable-Project%22-Templates) that are usable directly from source.

What you're adding here is totally fine for the `rel/2.0.x` branch, and thanks for sorting it out so quickly.

When we do this for real in `dev`, I'd love it if we could find a cleaner solution that keeps the build pipeline working in the expected ways, and without forcing devs to go through any unexpected extra steps. One possible approach would be *not* to replace these tokens at all during the package creation, so the resulting `.nupkg` contains `.csproj` files with dependencies expressed like this:

```
<PackageReference Include="Microsoft.AspNetCore.All" Version="$(TemplateDep_Coherence_2_1)" />
```

This would be runnable directly from source (and could be included in a regular `.sln`, etc.), because a `Directory.Build.props` file (or similar) would supply values based on the new `build/dependencies.props` you've added.

Then, when someone executes `dotnet new <template>`, we could have `dotnet new`'s template system resolve the symbols to fixed numbers. To do this we'd need entries in the `.template.config/template.json` files, adding to the `symbols` settings entries like `"TemplateDep_Coherence_2_1": { "replaces": "$(TemplateDep_Coherence_2_1)", "value": "2.1.2" }`.

So to make this work, we'd want our build system to generate or augment the `.template.config/template.json` files during the build. Previously you said it might be possible to modify the `.csproj` files in the staging directory during the build, so perhaps the same technique could be used instead to modify the `.template.config/template.json` files.

What do you think? This is just one possibility. If you have any other ideas that's great too!

natemcmaster: oooh, i like this. That would work well too. Didn't know this was a thing the templating system supported. Does the `.template.config/template.json` + `replaces` system work today? And btw, is there some documentation on available settings in template.json ?

SteveSandersonMS: Yes, it is supported today.

Docs for dotnet new's substitution features are all at [the link from before](https://github.com/dotnet/templating/wiki/%22Runnable-Project%22-Templates), or @sayedihashimi would know everything there is to know about it.

SteveSandersonMS:
> Template the templateconfig file instead

@natemcmaster Perfect! This must have been a laborious commit to put together, but great job - that looks like it gives us the best result both for the build system and for the developer experience!

natemcmaster:
:up: :date:

Changes
 - deleted .template.config/template.json files
 - added a bunch of parameters to symbols.parameters for package version replacement
 - dotnet-build produces the template.config/template.json files with hard-coded versions

natemcmaster:
Still trying to verify this actually works, of course....seeing some weird errors.

natemcmaster:
Found it. I was using template.json wrong. Everything is working smoothly now. :)

natemcmaster:
New error, but only on Linux/macOS
```
 ==> /Users/namc/.dotnet/dotnet new web [/Users/namc/dev/aspnet/templating/test/Templates.Test/bin/Debug/netcoreapp2.0/TestTemplates/37025cf1fb724909ad785b6ef06a34ac]
 [ERROR] Unexpected end when reading JSON. Path 'TemplateInfo[19].Tags.language.Description', line 1365, position 26.
 [ERROR]    at Newtonsoft.Json.JsonTextReader.ParseValue()
 [ERROR]    at Newtonsoft.Json.JsonTextReader.Read()
 [ERROR]    at Newtonsoft.Json.Linq.JContainer.ReadContentFrom(JsonReader r, JsonLoadSettings settings)
 [ERROR]    at Newtonsoft.Json.Linq.JContainer.ReadTokenFrom(JsonReader reader, JsonLoadSettings options)
 [ERROR]    at Newtonsoft.Json.Linq.JObject.Load(JsonReader reader, JsonLoadSettings settings)
 [ERROR]    at Newtonsoft.Json.Linq.JObject.Parse(String json, JsonLoadSettings settings)
 [ERROR]    at Newtonsoft.Json.Linq.JObject.Parse(String json)
 [ERROR]    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.EnsureTemplatesLoaded()
 [ERROR]    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.WriteTemplateCache(IList`1 templates, String locale)
 [ERROR]    at Microsoft.TemplateEngine.Edge.Settings.TemplateCache.WriteTemplateCacheForLocale(String locale, String existingCacheVersion)
 [ERROR]    at Microsoft.TemplateEngine.Edge.Settings.TemplateCache.WriteTemplateCaches(String existingCacheVersion)
 [ERROR]    at Microsoft.TemplateEngine.Edge.Settings.SettingsLoader.Save(TemplateCache cacheToSave)
```

Any ideas? I'm stumped on this one.

natemcmaster:
Open https://github.com/aspnet/templating/issues/63 to follow up on the test failure. The tests pass when parallelization is disabled.

natemcmaster:
FYI - 2.0.3 preview builds are up now. https://dotnet.myget.org/feed/aspnet-2-0-2-october2017-patch/package/nuget/Microsoft.AspNetCore.SpaTemplates

PR #128, opened 2017-10-26
[WIP] Adds HttpsMiddleware to templates

jkotalik:
Trying to see if AppVeyor triggers.

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates.2.1/content/RazorPagesWeb-CSharp/Startup.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates.2.1/content/RazorPagesWeb-CSharp/Startup.cs
@@ -122,6 +122,7 @@ namespace Company.WebApplication1
             else
             {
                 app.UseExceptionHandler("/Error");
+                app.UseHsts();
             }
 
             app.UseHttpsRedirection();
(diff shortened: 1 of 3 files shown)


src/Microsoft.DotNet.Web.ProjectTemplates.2.1/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -77,6 +78,11 @@ public void ConfigureServices(IServiceCollection services)
             .AddCookie();
 
 #endif
+            services.AddHsts(options =>
+            {
+                options.MaxAge = TimeSpan.FromDays(30);
+            });
+
```
javiercn: These lines can go away

src/Microsoft.DotNet.Web.ProjectTemplates.2.1/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -123,6 +124,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                 app.UseExceptionHandler("/Error");
             }
```
javiercn: We do want HSTS in all of these templates

PR #129, opened 2017-10-26
Generate the MSBuild files instead of the template.json files

natemcmaster:
Workaround https://github.com/dotnet/templating/issues/1302


SteveSandersonMS:
This is only for the 2.0.3 patch, and won't be needed in `dev`, is that right @natemcmaster?

natemcmaster:
I believe @mlorbetske has added a fix to the 2.1 templating engine for https://github.com/dotnet/templating/issues/1302. Unless that get's back-ported to the template engine in 2.0, we have to be cautious about this bug for any templates package that might be installed into a 2.0 CLI. So, maybe? 

@seancpeters @mlorbetske  is there a way for a template to require a minimum version of the SDK?

SteveSandersonMS:
I really don't want this change in `dev` so I definitely hope the 2.0.x CLI can be fixed. @seancpeters @mlorbetske

natemcmaster:
:+1: We should re-open https://github.com/dotnet/templating/issues/1302#issuecomment-339808277 (I don't have permissions on that repo)

mlorbetske:
I've replied to the issue on dotnet/templating. The short of it is unless we also update the engine in VS and VS for Mac and get partner sign off from Windows Template Studio and Xamarin, porting the engine fix to the 2.0.x CLI doesn't help with this issue.

mlorbetske:
Additionally, older versions of VS that look in the CLI location for templates would have the behavior we're looking to avoid. The engine fix must first be included in what we consider to be the oldest supported versions for consuming those templates before the change to use the new process can be used.

natemcmaster:
Ok, good to have the context. It sounds like it will be a while before we can use the constant generator then. @SteveSandersonMS I'll ping you again before any of this makes it back to dev.

PR #137, opened 2017-11-01
Adds Hsts and HttpsRedirection to spa templates

jkotalik:


src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Startup.cs
```diff
@@ -39,9 +40,11 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
             else
             {
                 app.UseExceptionHandler("/Home/Error");
+                app.UseHsts();
             }
 
             app.UseStaticFiles();
+            app.UseHttpsRedirection();
```
SteveSandersonMS: Could you clarify why `UseHttpsRedirection` goes after `UseStaticFiles`? Presumably this means that static files can be served over both `http` and `https`, but why is that desirable? Why not apply the redirection to requests for static files too?

javiercn: Good catch. This needs to go before

jkotalik: I'll check that all other templates are in the right order too.

PR #147, opened 2017-11-06
dev: Reorganize repo branch and project folder structure

natemcmaster:
When the code lived in dotnet/templating, the 1.x, 2.0, and 2.1 templates all needed to sit side-by-side.
Now that this code builds from aspnet/templating, we're reorganizing the folder and branch structure to
match the aspnetcore runtime repos.

Changes:
- Remove the 2.0 and 1.x templates projects from the `dev` branch. These will be serviced from rel/2.0.x and rel/1.1.x branches


src/Microsoft.DotNet.Web.ProjectTemplates/Microsoft.DotNet.Web.ProjectTemplates.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <PackageId>Microsoft.DotNet.Web.ProjectTemplates.2.1</PackageId>
     <NuspecFile>Microsoft.DotNet.Web.ProjectTemplates.2.1.nuspec</NuspecFile>
```
SteveSandersonMS: Could we also remove the version number from the nuspec filename (so future version updates will be less noisy)? Or does it have to match the PackageId?

natemcmaster: Good catch. File name can be version-less. I just missed this one. I'll update it.

PR #149, opened 2017-11-07
Update AngularSPA - use angular 5.0

NorDroN:


SteveSandersonMS:
Thanks very much for the contribution, @NorDroN!

As it happens, we're working on some major changes to the Angular template at the moment. The major thing is that it will become entirely based around Angular CLI. In the branch where we're building this, we've also migrated it to Angular 5 already.

Since we are getting the Angular 5 upgrade as part of a much wider package of changes, I'll close this. Hope that's OK, and thanks again for your involvement!

lehne:
@SteveSandersonMS I'm trying to set up my AspNetCore 2 app to use a live reload of the page. Webpack seems to be mostly working but Angular CLI with ng serve is faster and probably better in the long run.  Can I see the code for the template where you switch to the CLI? Also, does it use VS still or is it more VS Code focused?

Issue #150, opened 2017-11-10
Templates fail to compile


natemcmaster:
The followings tests currently fail with compiler errors:

Missing HttpsPolicy - webapi and mvc for .NET Framework
```
Failed   Templates.Test.MvcTemplateTest.MvcTemplate_IndividualAuth_Works_NetFramework
Error Message:
 System.Exception : Process exited with code 1
StdErr: 

Startup.cs(9,28): error CS0234: The type or namespace name 'HttpsPolicy' does not exist in the namespace 'Microsoft.AspNetCore' (are you missing an assembly reference?) [C:\b\w\5e6fce5f142a57e9\test\Templates.Test\bin\Release\netcoreapp2.0\TestTemplates\55db947774b64a4eb3897768078c7179\55db947774b64a4eb3897768078c7179.csproj]
```

F# templates
```
Failed   Templates.Test.MvcTemplateTest.MvcTemplate_NoAuth_Works_NetCore(languageOverride: "F#")
Error Message:
 System.Exception : Process exited with code 1
StdErr: 
StdOut: Microsoft (R) Build Engine version 15.5.172.19382 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

C:\b\w\5e6fce5f142a57e9\test\Templates.Test\bin\Release\netcoreapp2.0\TestTemplates\29409ff58d7e4c19b7cd4d7d48b48931\Program.fs(23,9): error FS0039: The value or constructor 'BuildWebHost' is not defined. Maybe you want one of the following:    Builder    IWebHost [C:\b\w\5e6fce5f142a57e9\test\Templates.Test\bin\Release\netcoreapp2.0\TestTemplates\29409ff58d7e4c19b7cd4d7d48b48931\29409ff58d7e4c19b7cd4d7d48b48931.fsproj]
```

http://aspnetci/viewLog.html?buildId=353915&tab=buildResultsDiv&buildTypeId=Lite_Templating

cc @jkotalik

jkotalik:
Should be fixed by #151

PR #151, opened 2017-11-10
Fix HttpsPolicy dependency for full framework

jkotalik:
For #150


natemcmaster:
Add MicrosoftAspNetCoreHttpsPolicyPackageVersion to dependencies.props too, then :shipit:

jkotalik:
Ah shoot, we need to update the tests. All tests are currently using http instead of https clients.

SteveSandersonMS:
@jkotalik Please ensure there's at least one test (per template) that still makes the initial request over `http` so we can verify that it correctly redirects over to `https`.

jkotalik:
@natemcmaster I'm going to merge this for now and fix the launch settings and Https tests in another PR (as this is a compiler blocker).

PR #157, opened 2017-11-16
Removes Https middlewares from templates

jkotalik:
In order to have the HTTPS middleware evaluated as a single PR, I am going to remove the HTTPS middlewares from the templates. The tests are currently failing on CI and it is best to evaluate HTTPS with all features completed, including tests, in a single PR. There is also uncertainty regarding whether we should have the HTTPS middleware in the templates.
cc/ @DamianEdwards @shirhatti

### DIFF
--- a/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/Startup.cs
+++ b/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/Startup.cs
@@ -1,7 +1,6 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.HttpsPolicy;
-using Microsoft.AspNetCore.SpaServices.Webpack;
+using Microsoft.AspNetCore.SpaServices.ReactDevelopmentServer;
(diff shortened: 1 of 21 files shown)


src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/Startup.cs
```diff
@@ -1,7 +1,7 @@
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.HttpsPolicy;
-using Microsoft.AspNetCore.SpaServices.ReactDevelopmentServer;
+using Microsoft.AspNetCore.SpaServices.Webpack;
```
Tratcher: ?

SteveSandersonMS: Well spotted. That change shouldn't be here, and I don't think the code will currently compile because of it (the extension method on L59 won't be found). Same problem in the ReactRedux template.

SteveSandersonMS:
@jkotalik Re-approved :)

Issue #92, opened 2017-10-24
Change "Web API" to "API"


Eilon:
_From @sayedihashimi on February 2, 2017 23:47_

We should change "Web API" to "API" in the One ASP.NET dialog.

Note: all credit goes to @blowdart for noticing this.

_Copied from original issue: aspnet/Templates#813_

PR #169, opened 2017-11-30
Change Web API to API

jbagga:
Addresses #92

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
@@ -3,10 +3,9 @@
   "author": "Microsoft",
   "classifications": [
     "Web",
-    "API",
     "WebAPI"
   ],
-  "name": "ASP.NET Core API",
+  "name": "ASP.NET Core Web API",
   "generatorVersions": "[1.0.0.0-*)",
   "description": "A project template for creating an ASP.NET Core application with an example Controller for a RESTful HTTP service. This template can also be used for ASP.NET Core MVC Views and Controllers.",
   "groupIdentity": "Microsoft.Web.WebApi",
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
@@ -3,9 +3,10 @@
   "author": "Microsoft",
   "classifications": [
     "Web",
+    "API",
     "WebAPI"
   ],
-  "name": "ASP.NET Core Web API",
+  "name": "ASP.NET Core API",
   "generatorVersions": "[1.0.0.0-*)",
   "description": "A project template for creating an ASP.NET Core application with an example Controller for a RESTful HTTP service. This template can also be used for ASP.NET Core MVC Views and Controllers.",
   "groupIdentity": "Microsoft.Web.WebApi",
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/vs-2017.3.host.json
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/vs-2017.3.host.json
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json.schemastore.org/vs-2017.3.host",
   "name": {
-    "text": "Web API",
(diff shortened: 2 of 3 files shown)


src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
```diff
@@ -3,9 +3,10 @@
   "author": "Microsoft",
   "classifications": [
     "Web",
+    "API",
     "WebAPI"
   ],
-  "name": "ASP.NET Core Web API",
```
Eilon: This is a product name, so I think it should be left as-is.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
```diff
@@ -3,9 +3,10 @@
   "author": "Microsoft",
   "classifications": [
     "Web",
+    "API",
```
Eilon: I'm not sure what a `classification` is - does it need to be changed?

jbagga: Is there a way I can see how these changes affect the dialog box? I thought it was like tags (for search results in the File->New dialog).

Eilon: Can we ask @seancpeters and/or @mlorbetske about it?

seancpeters: @phenning can probably give the best insight into how the classifications are used by Visual Studio.

phenning: We need to have a "web" classification for the items to show in one asp.net dialog, along with the host file.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/vs-2017.3.host.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "http://json.schemastore.org/vs-2017.3.host",
   "name": {
-    "text": "Web API",
+    "text": "API",
```
phenning: This won't change what gets displayed in the One ASP.NET dialog. The "text" value in the host file is only a fallback in case the package and resource id specified can not be loaded.

We have two options here:
1. I can make a resource change in the One ASP.NET Dialog to change the name of the template to "API". This would only apply to Dev 15.6 Preview 2+. This would impact both 1.x and 2.0 templates displayed,

2. Since in this case, API gets localized in all languages (I just confirmed in RUS and FRA),  you could simply remove the package and id properties here, in that case, in absence of those properties, VS will just use what is in the text field for all languages.

jbagga: @Eilon I like the first option. Just to be consistent with other templates and ensure that it's always localized correctly.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/template.json
jbagga: Okay, "API" is not required then. I have removed it from `classifications`

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/vs-2017.3.host.json
phenning: @Eilon if you are fine with this, I'll get this checked in for our next WTE build and insertion for 15.6.
@mlorbetske FYI

Eilon: Fine with me.

jbagga:
@Eilon I'll check this in as well so the fallback "text" field is API as well

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/.template.config/vs-2017.3.host.json
phenning: Thanks, will get this in the next build.

Issue #71, opened 2017-10-17
Unnecessary User.Id provided in password reset link?


tverboon:
The user Id is provided in the callback link:
https://github.com/aspnet/templating/blob/a80061f42be6ba7de36f8fea35086da58e08a69b/src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Controllers/AccountController.cs#L424

But in the controller action only code is used:
https://github.com/aspnet/templating/blob/a80061f42be6ba7de36f8fea35086da58e08a69b/src/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Controllers/AccountController.cs#L443

In my opinion it's better not to provide more information than necessary.

Copy of:  https://github.com/aspnet/Templates/issues/872

Eilon:
Indeed, it looks like it's not needed, so we should remove it.

PR #173, opened 2017-11-30
Remove user.Id from ResetPasswordCallbackLink

jbagga:
Addresses #71


javiercn:
I assume the User ID is not being used at all? /cc: @HaoK

jbagga:
@HaoK I'll merge this in for now. Let us know if you have any concerns.

PR #180, opened 2017-12-05
Set ErrorViewModel in ConfirmEmail

jbagga:
Addresses #80 

cc @Eilon


src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
```diff
@@ -346,7 +348,8 @@ public IActionResult ExternalLogin(string provider, string returnUrl = null)
                 throw new ApplicationException($"Unable to load user with ID '{userId}'.");
             }
             var result = await _userManager.ConfirmEmailAsync(user, code);
-            return View(result.Succeeded ? "ConfirmEmail" : "Error");
+            return View(result.Succeeded ? "ConfirmEmail" : 
+                "Error", new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
```
Eilon: I believe this will always pass in the model regardless of which view gets rendered. I think we want the error view model to be passed in only for the `Error` view.

jbagga:
@Eilon Refactored it

Issue #98, opened 2017-10-24
ExternalLoginConfirmation breaks if the username is already taken


Eilon:
_From @blowdart on January 18, 2017 20:18_

Copied from https://github.com/aspnet/Identity/issues/1080

Repro:

Login with a new Facebook account
In the "ExternalLoginConfirmation" page enter an email that already exists on the system
Press Register
The page reloads, the validation error is displayed saying "User name 'blah@example.com' is already taken." and all of the ViewData is now lost including the ReturnUrl of the form and the LoginProvider used in the h2 and intro paragraph.

_Copied from original issue: aspnet/Templates#793_

PR #187, opened 2017-12-09
Set LoginProvider when ExternalLoginConfirmation fails

jbagga:
Addresses #98

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
@@ -327,6 +327,8 @@ namespace Company.WebApplication1.Identity.Controllers
                         return RedirectToLocal(returnUrl);
                     }
                 }
+
+                ViewData["LoginProvider"] = info.LoginProvider;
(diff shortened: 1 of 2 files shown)


src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/AccountController.cs
```diff
@@ -327,6 +327,8 @@ public IActionResult ExternalLogin(string provider, string returnUrl = null)
                         return RedirectToLocal(returnUrl);
                     }
                 }
+
+                ViewData["LoginProvider"] = info.LoginProvider;
```
Eilon: Should this line go right before line 335 instead? That is, if ViewState is not valid, doesn't it also need this to be set?

I think you could test it out like so:
1. Disable JavaScript in the browser (easy to do using Chrome dev tools)
2. Navigate to the preceding page
3. Have a validation error in the form (whatever the form is)
4. Click the button that goes to this page

Eilon: And the same question applies to the Razor Pages code above.

jbagga:
@Eilon Updated

PR #188, opened 2017-12-10
Using nameof() instead of hard-coded string.

Biarity:


jbagga:
@Biarity Thank you for the PR and bringing this to our attention! Merging this in and working on changing other instances.

PR #196, opened 2017-12-13
Use nameof() for view names

jbagga:


jbagga:
@Eilon Waiting for it to pass on the CI. There has been some flakiness around one of the Selenium tests

PR #198, opened 2017-12-14
Prevent users from removing their last login if they don't have a password

Biarity:
While this is checked for in `GET /ExternalLogins` (by not showing the remove button), this is not confirmed in `POST /RemoveLogin`. This means a malicious user without a password can manually send a POST request to `/RemoveLogin` containing the details of their last login! This would result in having a password-less user in the database with no external logins.


Eilon:
Though we're not especially concerned about a malicious user attacking themselves, we'll consider this change. (A malicious user could also reset their password to something they don't remember and not be able to log in again anyway 😄 )

mkArtakMSFT:
Closing this PR as there was no activity here for quite a while now.

PR #199, opened 2017-12-14
Refresh user signin after updating their profile

Biarity:
That way, the changes are immediately visible.

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/ManageController.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Areas/Identity/Controllers/ManageController.cs
@@ -101,7 +101,8 @@ namespace Company.WebApplication1.Identity.Controllers
                     throw new ApplicationException($"Unexpected error occurred setting phone number for user with ID '{user.Id}'.");
                 }
             }
-
+            
+            await _signInManager.RefreshSignInAsync(user);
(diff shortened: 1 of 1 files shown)


Eilon:
I'm not sure if this is necessary because there's a redirect that's about to happen anyway. I feel that even though this API is called, the "system" is overall in a "bad state" because half of it will think the user is the "old" profile, and half of it will think the user is the "new" profile.

@HaoK / @Tratcher / @blowdart - thoughts on this?

Tratcher:
This makes sense. The user has just changed some of their profile information in the database but still has the old values cached as claims in the auth cookie. RefreshSignInAsync issues a new auth cookie with the updated information.

HaoK:
Its safe to call refresh sign in generally to update the claims on the response.  That said, identity doesn't put the phone number of email in the cookie by default, so this shouldn't actually result in any claims being updated

blowdart:
We should add it though, for "correctness" or an illustration of what should eb done.

Eilon:
@jbagga can you review & merge? @Tratcher and @blowdart say it's good, so let's do it.

jbagga:
@Eilon Which milestone is this in?

Eilon:
`2.1.0-preview2`, so just `dev`

jbagga:
The code has moved to Identity UI https://github.com/aspnet/Identity/pull/1615

jbagga:
@Biarity Thanks for the suggestion! The code change has been moved to the new location.

dougbu:
This issue was moved to aspnet/Scaffolding#736

PR #200, opened 2017-12-14
Target netcoreapp2.1

jbagga:


jbagga:
This is currently failing locally for just the Spa templates. I wanted to check if it passes on Appveyor while I try to figure out the cause of the error. It's a "Certification error"

jbagga:
@pranavkm FYI

pranavkm:
@javiercn do you know what the failure's about?

javiercn:
I'm relatively sure that's caused by edge not trusting the self-signed certificate. That said, trusting the certificate should make it go away or alternatively you can run the automation on firefox.

We can look this in more detail next time you are in the office. I don't know why it wasn't working and we didn't do enough diagnostics to ensure this was the case.

If the cert turns out to be issue we will automate clicking through the ignore ssl errors page on edge.

jbagga:
I guess one thing that will be interesting is if it can be reproduced locally on another machine. I just want to eliminate the possibility that this change to netcoreapp2.1 is responsible for the error.

javiercn:
Merged https://github.com/aspnet/templating/commit/886a9857262ac664b7fefe8b6fdfa29f063adc45

PR #168, opened 2017-11-30
Replace ApplicationException with IActionResult or meaningful exception

jbagga:
Addresses #9

PR #134, opened 2017-10-27
Update EF Core Migrations

bricelam:

PR #204, opened 2017-12-19
Replace identity code with built-in version on the templates

javiercn:
Updates the template to use the built-in UI from identity.
* Removes all the identity related code from the UI.
* Adds a reference for Microsoft.AspNetCore.Identity.UI into the csproj file.
* Updates Startup to use the built-in identity types.
* Updates the migrations to use the built-in identity types.

I grabbed @bricelam commit from https://github.com/aspnet/templating/pull/134 and the changes from https://github.com/aspnet/templating/pull/168 are part of the library with the embedded UI.


ajcvickers:
@javiercn I'm not ready to sign off on this until I see that the current issues with the initial generated migration have been fixed or mitigated, and I haven't heard anything about this. Do you know if this was done? The most important one is that the default index created is invalid on SQL Server. This didn't matter before because the migration didn't create the index. If it now does, then this somehow needs to be fixed in the Identity model, but it has not been clear to me yet how we can do this without breaking. /cc @HaoK @blowdart

javiercn:
@ajcvickers if you point me to the bug I’ll be happy to take a look at it and confirm. I cherry-picked @bricelam commit and put my changes on top. (And manually validated that the projects were able to create the database from scratch; for the templates we don’t care about ‘upgrading’ existing projects). If there is anything else to do here, let me know and I’ll happily take care of it. We can also update the migrations now and update them again later with the final fixes. This update just keeps the templates working.

javiercn:
After chatting with @ajcvickers about the migrations, I've taken out the update from @bricelam and just did the minimal update to the old migrations. (And verified that everything works). I will let @blowdart and @HaoK resolve the issue blocking the migrations update

I've normalized all the migrations to the version on the RazorPages template (the last one we shipped) and updated the one in the Starter web (wich was updated earlier this year to a preview version) to match the RazorPages version

PR #209, opened 2017-12-28
[WIP] Updates the templating repo to use netcoreapp2.1 and the new shared framework

javiercn:


build/repo.props
```diff
@@ -8,7 +8,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp20PackageVersion)" />
     <DotNetCoreRuntime Include="$(MicrosoftNETCoreApp21PackageVersion)" />
   </ItemGroup>
+
+  <ItemGroup>
+   <AspNetCoreRuntime Include="$(MicrosoftAspNetCoreAllPackageVersion)" />
```
natemcmaster: You probably need to also update dependencies.props. MicrosoftAspNetCoreAllPackageVersion is set to version before we started producing the shared framework.

javiercn: Yeah, just realized that. I'm adding that.

natemcmaster:
Hmm, we either need to fix this https://github.com/aspnet/Universe/pull/731, or update dotnet-install.ps1/sh to special case aspnetcore-runtime and download `win7-x64` instead of `win-x64`.

javiercn:
@natemcmaster I've signed-off your PR to fix it. We should update universe IMO and have everything be consistent

javiercn:
Done

Issue #215, opened 2017-12-31
Routes using query string parameters don't work in the Knockout-CSharp template


Bumbolio:
I attempted to register a route that had a query string parameter but crossroads wasn't picking it up.  Here is an example route.  
```
{ url: 'linkaccount{?query}', params: { page: 'streaming-services' } },
```
The problem was that the query string portion of the url was not included when invoking `crossroads.parse()`.  By including `document.location.search` in the string being passed to `crossroads.parse()`, this issue is resolved.  Followed the History.js example from the [Crossroads documentation](https://github.com/millermedeiros/crossroads.js/wiki/Examples#using-with-historyjs). 

Example fix:
```
crossroads.parse(history.location.pathname + document.location.search);
```
```
this.disposeHistory = history.listen(location => crossroads.parse(location.pathname + document.location.search));
```

I can send a pull request if needed, but I couldn't reach the site to sign up as a contributor (https://cla2.dotnetfoundation.org/).

Bumbolio:
I went a head and opened PR #216 with a proposed fix.

SteveSandersonMS:
See #216 for info

PR #216, opened 2017-12-31
Added support for query string parameters in Knockout-CSharp template.

Bumbolio:
Summary of the changes (Less than 80 chars)
 - Updated the initial route for Crossroads so that it includes the query string portion of the URL.
 - Updated the handler for browser history changes so that it includes the query string portion of the URL.  

Addresses #215

Information:

I attempted to register a route that had a query string parameter but crossroads wasn't picking it up.  Here is an example route.  
```
{ url: 'linkaccount{?query}', params: { page: 'streaming-services' } },
```
The problem was that the query string portion of the url was not included when invoking `crossroads.parse()`.  By including `document.location.search` in the string being passed to `crossroads.parse()`, this issue is resolved.  Followed the History.js example from the Crossroads documentation.  https://github.com/millermedeiros/crossroads.js/wiki/Examples#using-with-historyjs

### DIFF
--- a/src/Microsoft.AspNetCore.SpaTemplates/content/Knockout-CSharp/ClientApp/router.ts
+++ b/src/Microsoft.AspNetCore.SpaTemplates/content/Knockout-CSharp/ClientApp/router.ts
@@ -28,7 +28,7 @@ export class Router {
         });
 
         // Make history.js watch for navigation and notify Crossroads
-        this.disposeHistory = history.listen(location => crossroads.parse(location.pathname));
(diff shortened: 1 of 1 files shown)


Eilon:
@SteveSandersonMS - thoughts on this?

SteveSandersonMS:
Thanks @Bumbolio for proposing this!

We're now recommending that interested community members [publish their own Knockout templates separately](https://docs.microsoft.com/en-us/dotnet/core/tools/custom-templates). The `dotnet new` system is extensible for this exact reason. We're no longer updating the Knockout template in this repo ([details](https://github.com/aspnet/Announcements/issues/289)).

PR #219, opened 2018-01-04
Add HttpsPort settings to Empty F# template

jbagga:


jbagga:
Web API and Starter Web templates for F# already has these settings https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/.template.config/template.json#L46

Issue #45, opened 2017-10-04
Update Web API templates to use new Web API conventions 


danroth27:
We are adding some new [Web API specific conventions]() that we will need the Web API project and time templates to react to.

We are adding `[ApiController]` attribute as the way to opt-in to new API specific conventions and behaviors that are tuned around best practices for APIs.

This will include:
- Automatically responding with a 400 when validation errors occur (using https://tools.ietf.org/html/rfc7807)
- Infer smarter defaults for action parameters (`[FromBody]` for complex types, `[FromRoute]` when possible, otherwise `[FromQuery]`)
- Web API controllers should derive from ControllerBase (see https://github.com/aspnet/templating/issues/82)

Eilon:
Added this to the list above:

- Web API controllers should derive from ControllerBase (see https://github.com/aspnet/templating/issues/82)

Eilon:
@danroth27 / @rynowak / @pranavkm - should we get this into Preview1? Seems like a simple change that will help us get feedback on the greatness of `[ApiController]` and friends

pranavkm:
With the exception of the first bullet not being fully fleshed out (we have the validation filter, we don't do RFC7807), there's enough in here to start getting some feedback.

rynowak:
Should we add `[ApiController]`?

danroth27:
I think we should

PR #220, opened 2018-01-05
Derive Web API templates from ControllerBase

jbagga:
Addresses part of #45 

cc @Eilon

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
@@ -13,7 +13,7 @@ namespace Company.WebApplication1.Controllers
     [Authorize]
 #endif
     [Route("api/[controller]")]
-    public class ValuesController : Controller
+    public class ValuesController : ControllerBase
     {
         // GET api/values
         [HttpGet]
@@ -24,7 +24,7 @@ namespace Company.WebApplication1.Controllers
 
         // GET api/values/5
         [HttpGet("{id}")]
-        public string Get(int id)
+        public string Get([FromRoute]int id)
         {
             return "value";
         }
@@ -40,7 +40,7 @@ namespace Company.WebApplication1.Controllers
 
         // PUT api/values/5
         [HttpPut("{id}")]
-        public void Put(int id, [FromBody]string value)
+        public void Put([FromRoute]int id, [FromBody]string value)
         {
 #if (OrganizationalAuth || WindowsAuth)
             // For more information on protecting this API from Cross Site Request Forgery (CSRF) attacks, see https://go.microsoft.com/fwlink/?LinkID=717803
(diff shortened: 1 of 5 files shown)


src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
```diff
@@ -24,7 +24,7 @@ public IEnumerable<string> Get()
 
         // GET api/values/5
         [HttpGet("{id}")]
-        public string Get(int id)
+        public string Get([FromRoute]int id)
```
rynowak: Space between `[FromRoute]` and `int id`

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
```diff
@@ -13,7 +13,7 @@ namespace Company.WebApplication1.Controllers
     [Authorize]
 #endif
     [Route("api/[controller]")]
-    public class ValuesController : Controller
+    public class ValuesController : ControllerBase
```
pranavkm: Need to annotate this type with `ApiController`

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
pranavkm: Shouldn't need this, we should infer that it's `FromRoute` because it appears as a route value.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
jbagga: Is that all we are doing for this PR then? Derive from ControllerBase and annotate with ApiController? Since FromRoute isn't required.

jbagga:
@pranavkm @rynowak Updated

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
```diff
@@ -1,14 +1,11 @@
 ﻿namespace Company.WebApplication1.Controllers
 
-open System
```
pranavkm: Why remove these?

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
pranavkm: Essentially given ValuesController isn't very complicated.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
jbagga: Unused

pranavkm: We intentionally keep commonly used usings here. Let's undo this bit.

jbagga: ok

Issue #62, opened 2017-10-12
Make sure EF migrations in all templates are up-to-date with latest EF migrations code


Eilon:
cc @ajcvickers FYI

Eilon:
cc @HaoK @javiercn because of the discussions we're having.

Eilon:
@ajcvickers I'm assigning this to you as well because it's tracking https://github.com/aspnet/templating/pull/134.

Eilon:
And this is listed in preview1, so let's find out ASAP what needs to be done.

javiercn:
I believe this is blocked on https://github.com/aspnet/Identity/issues/1497. I'll let @ajcvickers and @HaoK  and @blowdart figure it out.

javiercn:
The main issue is that the updated migrations apparently create indexes and those indexes are too big in some platforms, but I don't have all the context.

ajcvickers:
@HaoK and I came up with a plan and he is working on it as a priority. Shouldn't be more than a few days work.

HaoK:
5d2f598c8094f4c0b00e430ddd50512fc32fe3eb

PR #227, opened 2018-01-09
Turn on MaxKeyLengthForKeys in templates

HaoK:
Depends on: https://github.com/aspnet/Identity/pull/1555

Part of: https://github.com/aspnet/templating/issues/62

@javiercn @ajcvickers


javiercn:
@ajcvickers Shouldn't we update the migrations at the same time?

ajcvickers:
@javiercn The migration needs to be updated, but it doesn't have to be done as part of the same PR.

javiercn:
@ajcvickers Won't this change break dotnet new?

ajcvickers:
@javiercn Break in what way?

javiercn:
@ajcvickers Won't ef tell you that the model is not up to date and you need to generate a new migration?

ajcvickers:
@javiercn Essentially, no. More specifically:
* EF Core doesn't have any mechanism that runs automatically for checking if the model is up-to-date with the database.
  * There is the database error page middleware, but that only kicks in if the application fails. The application should not fail in this case unless it has data that won't fit, which is unlikely, and is an error that would happen later anyway.
* If the user calls add-migration, then a new migration will be scaffolded. But this is true anyway at the moment because the existing migration checked in is not up-to-date anyway.

HaoK:
Updated PR with initial migration updates as well, please re-review

HaoK:
5d2f598c8094f4c0b00e430ddd50512fc32fe3eb

PR #244, opened 2018-01-19
Add firewall exclusions for test executables

jbagga:
Proposed workaround for the Windows Firewall prompts that appear when running templating tests locally.

It requires running tests in a process running as the Administrator as elevated privileges are required to add firewall exclusions. If not run as an Admin, the prompts will still appear and the tests will not be interrupted otherwise (will not fail because of these changes).

cc @mkArtakMSFT

### DIFF
--- a/test/Templates.Test/Helpers/AddFirewallExclusion.cs
+++ b/test/Templates.Test/Helpers/AddFirewallExclusion.cs
@@ -1,0 +1,70 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace Templates.Test.Helpers
+{
+    internal class AddFirewallExclusion : IDisposable
+    {
+        private bool _disposedValue = false;
+        private readonly string _exclusionPath;
+
+        public static IDisposable Create(string exclusionPath)
+        {
+            return new AddFirewallExclusion(exclusionPath);
+        }
+
+        private AddFirewallExclusion(string exclusionPath)
+        {
+            if (!File.Exists(exclusionPath))
+            {
+                throw new FileNotFoundException();
+            }
+
+            _exclusionPath = exclusionPath;
+            var startInfo = new ProcessStartInfo
+            {
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+                CreateNoWindow = true,
+                FileName = "cmd.exe",
+                Arguments = $"/c netsh advfirewall firewall add rule name=\"Allow {exclusionPath}\" dir=in action=allow program=\"{exclusionPath}\"",
+                UseShellExecute = false,
+                Verb = "runas",
+                WindowStyle = ProcessWindowStyle.Hidden,
+            };
+
+            Process.Start(startInfo);
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (!_disposedValue)
+            {
+                if (disposing)
+                {
+                    var startInfo = new ProcessStartInfo
+                    {
+                        RedirectStandardError = true,
+                        RedirectStandardOutput = true,
+                        CreateNoWindow = true,
+                        FileName = "cmd.exe",
+                        Arguments = $"/c netsh advfirewall firewall delete rule name=\"Allow {_exclusionPath}\"",
+                        UseShellExecute = false,
+                        Verb = "runas",
+                        WindowStyle = ProcessWindowStyle.Hidden,
+                    };
+
+                    Process.Start(startInfo);
+                }
+
+                _disposedValue = true;
+            }
+        }
+    }
+}
--- a/test/Templates.Test/Helpers/AddFirewallExclusion.cs
+++ b/test/Templates.Test/Helpers/AddFirewallExclusion.cs
@@ -9,16 +9,11 @@ namespace Templates.Test.Helpers
         private bool _disposedValue = false;
         private readonly string _exclusionPath;
 
-        public static IDisposable Create(string exclusionPath)
-        {
-            return new AddFirewallExclusion(exclusionPath);
-        }
-
-        private AddFirewallExclusion(string exclusionPath)
+        public AddFirewallExclusion(string exclusionPath)
         {
             if (!File.Exists(exclusionPath))
             {
-                throw new FileNotFoundException();
+                throw new FileNotFoundException($"File {exclusionPath} was not found.");
             }
 
             _exclusionPath = exclusionPath;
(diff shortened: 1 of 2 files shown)


jbagga:
We can discuss this and other tweaks that can be made to the tests next week @SteveSandersonMS

test/Templates.Test/Helpers/AddFirewallExclusion.cs
```diff
@@ -0,0 +1,70 @@
...
+        public static IDisposable Create(string exclusionPath)
+        {
+            return new AddFirewallExclusion(exclusionPath);
+        }
+
+        private AddFirewallExclusion(string exclusionPath)
+        {
+            if (!File.Exists(exclusionPath))
+            {
+                throw new FileNotFoundException();
```
SteveSandersonMS: Very minor point, but we could pass `exclusionPath` to the `FileNotFoundException` constructor so it gives a clearer error message.

test/Templates.Test/Helpers/AddFirewallExclusion.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            _exclusionPath = exclusionPath;
+            var startInfo = new ProcessStartInfo
+            {
+                RedirectStandardError = true,
+                RedirectStandardOutput = true,
+                CreateNoWindow = true,
+                FileName = "cmd.exe",
+                Arguments = $"/c netsh advfirewall firewall add rule name=\"Allow {exclusionPath}\" dir=in action=allow program=\"{exclusionPath}\"",
```
SteveSandersonMS: Out of interest, does this require you to be running as Administrator? Do our tests already fail if you're not Administrator?

test/Templates.Test/Helpers/AspNetProcess.cs
```diff
@@ -74,15 +74,26 @@ public AspNetProcess(ITestOutputHelper output, string workingDirectory, string p
...
             }
             else
             {
                 var exeFullPath = publish
                     ? Path.Combine(workingDirectory, $"{projectName}.exe")
                     : Path.Combine(workingDirectory, "bin", "Debug", framework, $"{projectName}.exe");
-                _process = ProcessEx.Run(output, workingDirectory, exeFullPath, envVars: envVars);
+                using (AddFirewallExclusion.Create(exeFullPath))
+                {
+                    _process = ProcessEx.Run(output, workingDirectory, exeFullPath, envVars: envVars);
```
SteveSandersonMS: Assuming the firewall exclusion only needs to exist for the period where we're *starting* up the `.exe` and it *begins* the TCP listener, this looks great.

test/Templates.Test/Helpers/AddFirewallExclusion.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.IO;
+
+namespace Templates.Test.Helpers
+{
+    internal class AddFirewallExclusion : IDisposable
+    {
+        private bool _disposedValue = false;
+        private readonly string _exclusionPath;
+
+        public static IDisposable Create(string exclusionPath)
```
SteveSandersonMS: Based on the usage below, it looks like this class could just have a regular `public` constructor instead of a `static` factory function. Not a big deal - I'm not fixated on whether this is changed or not.

test/Templates.Test/Helpers/AddFirewallExclusion.cs
SteveSandersonMS: Or to be clearer, I suppose the existence of this code implies our tests aren't already running as Administrator, because if they were then this would not be needed.

My only concern would be if this could sometimes pop up the Windows UAC dialog but I suppose it must not do because otherwise the tests would always fail.

test/Templates.Test/Helpers/AspNetProcess.cs
```diff
@@ -74,15 +74,26 @@ public AspNetProcess(ITestOutputHelper output, string workingDirectory, string p
             {
                 var dllPath = publish ? $"{projectName}.dll" : $"bin/Debug/{framework}/{projectName}.dll";
                 _process = ProcessEx.Run(output, workingDirectory, DotNetMuxer.MuxerPathOrDefault(), $"exec {dllPath}", envVars: envVars);
+                var listeningUrlString = GetListeningUrlString(output);
+                _listeningUri = new Uri(listeningUrlString, UriKind.Absolute);
```
SteveSandersonMS: Since in both cases where `GetListeningUrlString` is called, we immediately wrap the return value in a `new Uri`, would it be cleaner to have `GetListeningUrlString` itself return the `Uri` instance? (Presumably also rename it to something like `GetListeningUri`).

test/Templates.Test/Helpers/AddFirewallExclusion.cs
SteveSandersonMS: OK ignore these comments. I see from your initial PR description how this works. Looks good.

jbagga: Yes it requires you to be running the process as an admin (either in VS or cmd line) and we don't do that in our CI builds. But that's okay since the existence of this code does not lead to failures if not running as admin. This is just useful when building locally.

Question for you @SteveSandersonMS how come building release/2.0 branch does not have the firewall prompts? What changed since then that is causing the firewall warnings in the first place?

SteveSandersonMS: I'm not aware of anything we've changed in our source code that would affect this. Does it definitely *not* produce the prompts now if we build `release/2.0` still now? I thought maybe the CI machines have changed in some way that has reset whatever TCP-listening permission it needs.

jbagga: I meant while building locally. It does not show the prompts for release/2.0 at all.

jbagga:
Updated

test/Templates.Test/Helpers/AddFirewallExclusion.cs
```diff
@@ -9,16 +9,11 @@ internal class AddFirewallExclusion : IDisposable
         private bool _disposedValue = false;
         private readonly string _exclusionPath;
 
-        public static IDisposable Create(string exclusionPath)
-        {
-            return new AddFirewallExclusion(exclusionPath);
-        }
-
-        private AddFirewallExclusion(string exclusionPath)
+        public AddFirewallExclusion(string exclusionPath)
```
SteveSandersonMS: With this change to the API, there's a risk that `AddFirewallExclusion` could behave differently than a developer might expect. Since the method is called `Add...`, it suggests that you can call it multiple times and add multiple firewall exclusions. But if you do that, then `Dispose` won't do what you'd expect (it will only remove the last one you added). I'd suggest that `exclusionPath` should only be passed to the constructor, and not modifyable afterwards. In fact you probably only need a constructor, and don't need a separate `AddFirewallExclusion` call.

SteveSandersonMS:
Looks great!

BTW I posted a completely bogus suggestion about the code a minute ago, then immediately realised I was wrong. So if you got an email notification about `AddFirewallExclusion` just now, please ignore it.

PR #246, opened 2018-01-22
Update Identity to use the default layout from the template

javiercn:


javiercn:
Stuff is failing on the CI due to something wrong with the CLI we are consuming, but I verified the changes manually on full framework.

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Areas/Identity/Pages/_ViewStart.cshtml
```diff
@@ -0,0 +1,3 @@
+﻿@{
+    Layout = "/Pages/_Layout.cshtml";
```
DamianEdwards: When I tried this it wasn't enough, as the manage pages use a nested layout page. Overriding it required copying that layout page into the app, which is less than ideal of course. We should have a think/chat about how we could make this better, if possible.

javiercn: This is fixed now https://github.com/aspnet/Identity/commit/df4db2f49c80ef310067af0808b04b0acf87e148

For other Razor libraries with UI we just need provide guidance to our customers on how to do this. Esentially it comes down to not relying on hardcoded paths.

PR #249, opened 2018-01-22
Add application DbContext back to the templates

javiercn:


src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -80,24 +76,15 @@ public void ConfigureServices(IServiceCollection services)
...
-    #elif (IndividualB2CAuth)
+#elif (IndividualB2CAuth)
                 .AddAzureAdB2C(options => Configuration.Bind("AzureAdB2C", options))
-    #endif
+#endif
             .AddCookie();
 
 #endif
-#if (IndividualLocalAuth)
-            services.AddMvc()
```
javiercn: This is just extra clean-up

PR #251, opened 2018-01-23
Add IE Polyfills to the Angular SPA Template

DustinEwers:
Uncommented the IE polyfills in the Angluar SPA template

Reasoning: 
Most people who are building ASP.NET Core applications are going to be supporting IE. Defaulting to support IE would make the template more useful.


Eilon:
Hi @DustinEwers , it was an explicit decision to not include the polyfills by default because they are rarely needed. For the apps that need it, we tried to make it very easy to enable.

cc @SteveSandersonMS

DustinEwers:
Interesting. I’ve needed these on every Angular app I’ve built. Regardless, it’s your call. Thanks for the template and responding to this request. I’m really enjoying it thus far.

Eilon:
@DustinEwers sure thing. From the general consensus we've seen, and accounting for browser market shares, we felt the benefit wasn't worth the cost. Glad you're enjoying it!

PR #260, opened 2018-01-25
Remove RuntimeFrameworkVersion from the templates

natemcmaster:
This property is no longer needed because the Microsoft.AspNetCore.All and .App packages set the shared framework version. RuntimeFrameworkVersion is inferred by Microsoft.NET.Sdk.

Note: this targets <kbd>release/2.1</kbd>

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,7 @@
 ﻿init:
   - git config --global core.autocrlf true
 install:
-   - ps: Install-Product node 8.4.0 x64
+   - ps: Install-Product node 8 x64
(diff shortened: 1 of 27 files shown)


pranavkm:
:shipit: once build passes

natemcmaster:
Tests are failing with what appear to be unrelated errors. https://github.com/aspnet/templating/issues/261  - @SteveSandersonMS is this a known issue in release/2.1 that we can ignore?

natemcmaster:
Upgrading CI to use Node 8

Issue #263, opened 2018-01-26
Update ASP.NET Core 2.1 templates to use fixed ports when running from the command-line


danroth27:
The ASP.NET Core 2.1 templates currently generate random ports for the generated app, including the project launch profile. This make it tedious to then browse to the application when launched from the command line. We should change the project launch profile to specify explicitly the default ports (5000 & 5001) using the ASPNETCORE_URLS environment variable.

Eilon:
@jbagga here's a preview1 item for you.

Eilon:
Please sync with @danroth27 if you have questions on this.

PR #264, opened 2018-01-27
Set ASPNETCORE_URLS

jbagga:
Addresses #263 

cc @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
+++ b/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
@@ -21,7 +21,7 @@
       "launchBrowser": true,
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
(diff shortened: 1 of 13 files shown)


src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
```diff
@@ -21,7 +21,7 @@
       "launchBrowser": true,
       "environmentVariables": {
         "ASPNETCORE_ENVIRONMENT": "Development",
-        "ASPNETCORE_URLS": "https://localhost:44300;http://localhost:8080"
+        "ASPNETCORE_URLS": "https://localhost:5000;http://localhost:5001"
```
kichalla: If these ports are typically used for other purposes(on different OSes), then we might want to change them to something else.

Eilon: These are not reserved ports. They're commonly used by various web frameworks for dev.

PR #300, opened 2018-02-04
:lipstick: Remove unused ID attributes

peterblazejewicz:
This is jest re-brushing simple commit to trim down content.
These specific ids have no purpose behind them. For example these forms are not
validated explicitly by client code (during validation), there is no CSS style
matched for them, etc.

Thanks!

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Pages/Shared/_LoginPartial.cshtml
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Pages/Shared/_LoginPartial.cshtml
@@ -5,7 +5,7 @@
 
 @if (SignInManager.IsSignedIn(User))
 {
-    <form asp-area="Identity" asp-page="/Account/Logout" asp-route-returnUrl="@Url.Page("/Index", new { area = "" })" method="post" id="logoutForm" class="navbar-right">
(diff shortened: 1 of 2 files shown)


Eilon:
Hi @peterblazejewicz I think these are useful for testing, even if they're not used directly in the template. They also make it more convenient to style things later by just adding CSS. So, I think given that it's fairly harmless, and potentially useful, it's ok to keep them. Thanks!

peterblazejewicz:
@Eilon 
Yes, the QA ids is a valid point,

For above reason:
note to myself, for the records:
- find and add ids to all those forms without IDs (mvc with auth):
```cshtml
<form asp-action="ForgotPassword" method="post">
<form asp-route-returnurl="@ViewData["ReturnUrl"]" method="post">
<form method="post" asp-route-returnUrl="@ViewData["ReturnUrl"]">
<form asp-action="ExternalLoginConfirmation" asp-route-returnurl="@ViewData["ReturnUrl"]" method="post">
```

peterblazejewicz:
@Eilon 
Bear with me, but where I could find now the templates (Razor) that comes as part of shared content for `ASP.NET MVC` templates? Kind of older:
https://github.com/aspnet/Templates/tree/dev/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Views/Manage
TIA

Eilon:
@peterblazejewicz for the 2.1 release we moved almost all the "individual authentication" code and markup into a library NUPKG.

Here's the work item: https://github.com/aspnet/Identity/issues/1473
And this is where the code lives now: https://github.com/aspnet/Identity/tree/dev/src/UI

peterblazejewicz:
Thx!

PR #308, opened 2018-02-07
Update initial schemas

HaoK:
@ajcvickers @javiercn 

This seems pretty odd behavior, but the ordering of the properties seems to be the only diff, but that causes add-migration to generate add/drops...


HaoK:
PR targeting 2.1 is here: https://github.com/aspnet/templating/pull/311

PR #311, opened 2018-02-07
[2.1] Update initial migration

HaoK:
Migrations appear to be affected by ordering of properties (not sure if this is an underlying migration bug)

@ajcvickers


HaoK:
Replaced by https://github.com/aspnet/templating/pull/313

PR #313, opened 2018-02-07
[2.1] Update model snapshot

HaoK:
New update for initial migrations


HaoK:
Rather than cherry picking changes, instead take a fresh snapshot of all 3 migration files.  

One last issue is whether we can get rid of the MaxLength ifdef since Add-Migration in VS seems to be generating the HasMaxLength code now

HaoK:
@ajcvickers verified things work as expected with @bricelam and removed the ifdefs for MaxLength.  I also tried the full scaffold/AddMovie flow in the original devdiv bug which now works and generates only migrations related to the new movie table as expected.

Issue #309, opened 2018-02-07
Switch from checked in initial migration to running Add-Migration as part of build


HaoK:
Currently the migrations are constantly getting stale and broken.  We should look into switching to generating the initial migration during project creation since its just running the "Add-Migration Initial" command.  This guarantees the content will be fresh unlike today where the migration will just be broken if it becomes stale.

HaoK:
Met with @seancpeters @Eilon @javiercn yesterday and discussed with @ajcvickers today.

We agreed that the best long term plan is for the templating repo to produce the migrations during build.

Rough thinking is to:
1. Build the templates like today.
2. Instantiate each flavor of the template via `dotnet new razor -au Individual` and one with `--use_local_db` specifying the project name to match the expected format "Company.WebApplication1"
3. Apply any fixup needed (maybe nothing)
4. copy the generated migrations into the matching Razor/StarterWeb.CSharp locations so they will be updated automatically.

Eilon:
@HaoK can you take a crack at this? I chatted with @ajcvickers and he said you probably can fit this in.

HaoK:
Looks like we've come full circle, I spent some time with @natemcmaster trying to figure out how we would do this...  The problem with build time, is the conditional logic between SqlServer/SqlLite, even if we generate the correct migration for each, we'd have to merge them back somehow into a single one.  This was why the runtime approach where we created the migrations after dotnet new was better, since it eliminates the need for this switch...

Anyways, I think the cheapest and highest ROI at this point is for me to just create a test in templating that does:

dotnet new (with and without sqlserver), and then runs dotnet ef migrations and verifies that the new migration is empty.

Sound reasonable?  Basically we'd just have a CI failure whenever migrations need to be updated instead of automatically updating them.

ajcvickers:
@HaoK That sounds reasonable to me. @Eilon?

javiercn:
> The problem with build time, is the conditional logic between SqlServer/SqlLite, even if we generate the correct migration for each, we'd have to merge them back somehow into a single one.

I believe this might be possible. You can have them be
Data/Migrations/SqlLite/... and Data/Migrations/SqlServer/... in the template content and move them to Data/Migrations/... when we instantiate the template.

@mlorbetske Can you teach us the ways of the template engine?

HaoK:
[...]

PR #344, opened 2018-03-06
Add test for ensuring that migration is current

HaoK:
Fixes https://github.com/aspnet/templating/issues/309#issuecomment-370951849

cc @ajcvickers @Eilon @javiercn @natemcmaster

### DIFF
--- a/test/Templates.Test/Helpers/TemplateTestBase.cs
+++ b/test/Templates.Test/Helpers/TemplateTestBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Threading;
 using Microsoft.Extensions.CommandLineUtils;
@@ -79,6 +80,18 @@ namespace Templates.Test
             }
         }
 
+        protected void RunDotNetEfCreateMigration(string migrationName)
+        {
+            var args = $"ef migrations add {migrationName}";
+
+            // Only run one instance of 'dotnet new' at once, as a workaround for
+            // https://github.com/aspnet/templating/issues/63
+            lock (DotNetNewLock)
+            {
+                ProcessEx.Run(Output, TemplateOutputDir, DotNetMuxer.MuxerPathOrDefault(), args).WaitForExit(assertSuccess: true);
+            }
+        }
+
         protected void AssertDirectoryExists(string path, bool shouldExist)
         {
             var fullPath = Path.Combine(TemplateOutputDir, path);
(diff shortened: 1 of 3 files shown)


test/Templates.Test/Helpers/TemplateTestBase.cs
```diff
@@ -94,6 +107,26 @@ protected void AssertDirectoryExists(string path, bool shouldExist)
...
+            var emptyMigration = @"protected override void Up(MigrationBuilder migrationBuilder)
+        {
+
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+
+        }";
+            Assert.Contains(emptyMigration, contents);
```
javiercn: Don't you have to check other files? Or is the idea that if something fails we'll just regenerate everything under Data/Migrations

test/Templates.Test/MvcTemplateTest.cs
```diff
@@ -86,6 +86,10 @@ private void MvcTemplate_IndividualAuthImpl(string targetFrameworkOverride, bool
             Assert.Contains("Microsoft.VisualStudio.Web.CodeGeneration.Design", projectFileContents);
             Assert.Contains("Microsoft.EntityFrameworkCore.Tools.DotNet", projectFileContents);
 
+            RunDotNetEfCreateMigration("mvc");
+
+            AssertEmptyMigration("mvc");
+
```
javiercn: Should we also try to do dotnet ef database update here to ensure that the migration is applied correctly?

ajcvickers: Probably not needed here if the generated migration is really empty.

test/Templates.Test/Helpers/TemplateTestBase.cs
ajcvickers: It's not really required, but we could try to verify the snapshot. But I'm not sure it is worth it since I can't think of a way that it could be wrong without the actual generated migration being non-empty,

test/Templates.Test/Helpers/TemplateTestBase.cs
```diff
@@ -94,6 +107,26 @@ protected void AssertDirectoryExists(string path, bool shouldExist)
...
 
+        protected void AssertEmptyMigration(string migration)
+        {
+            var fullPath = Path.Combine(TemplateOutputDir, "Data/Migrations");
+            var file = Directory.EnumerateFiles(fullPath).Where(f => f.EndsWith($"{migration}.cs")).FirstOrDefault();
+
+            Assert.NotNull(file);
+            var contents = File.ReadAllText(file);
+
+            var emptyMigration = @"protected override void Up(MigrationBuilder migrationBuilder)
```
ajcvickers: Is this going to work when run on Linux? Will there be new-line issues due to differences in what is in the source file (not generated on Linux) and what is in the generated file?

test/Templates.Test/Helpers/TemplateTestBase.cs
javiercn: Yeah, that was my point. I just wanted to confirm

test/Templates.Test/Helpers/TemplateTestBase.cs
HaoK: Doesn't look like we need to worry about linux/mac, the top level tests are attributed like so for individual auth:

```
        [ConditionalFact]
        [OSSkipCondition(OperatingSystems.Linux)]
        [OSSkipCondition(OperatingSystems.MacOSX)]```

test/Templates.Test/Helpers/TemplateTestBase.cs
HaoK: Yeah the main idea is just to have tripwire so we have a failure

HaoK: If we ever end up with a situation where something in the migration is wrong and the test didn't fail, we can update the test with that additional case, but I don't think its worth spending too much time making this any more complicated up front

PR #348, opened 2018-03-08
Split sqllite and sql server migrations

HaoK:
Part of https://github.com/aspnet/templating/issues/309

@seancpeters does this look right?  We already have a UseLocalDb symbol we can use...  I tried this and I don't think the rename/exclusion is working as I have it, since both sets of files are showing up still...Can you check if there something obviously wrong?


seancpeters:
@HaoK - I believe that the problem is because template symbols are case-sensitive. In the symbols section, there is UseLocalDB, but in the source modifier conditions it's UseLocalDb.

HaoK:
@seancpeters that was it!  After talking with @ajcvickers I moved things around a bit, so there's now a Data\SqlLite and Data\SqlServer which contain the 3 migration files, and they get renamed to Data\Migrations.  

This should be ready for review now, the functional mvc/razor tests are passing

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
```diff
@@ -12,20 +12,66 @@
...
+                        .HasMaxLength(256);
+
+                    b.Property<string>("NormalizedName")
+                        .HasMaxLength(256);
+
+                    b.HasKey("Id");
+
+                    b.HasIndex("NormalizedName")
+                        .IsUnique()
+#if (UseLocalDB)
```
ajcvickers: Shouldn't need this anymore, right?

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlServer/ApplicationDbContextModelSnapshot.cs
```diff
@@ -12,20 +12,66 @@
...
+                        .HasMaxLength(256);
+
+                    b.Property<string>("NormalizedName")
+                        .HasMaxLength(256);
+
+                    b.HasKey("Id");
+
+                    b.HasIndex("NormalizedName")
+                        .IsUnique()
+#if (UseLocalDB)
```
ajcvickers: Likewise.

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
HaoK: Oops yeah good catch I'll remove these two

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
```diff
@@ -12,20 +12,66 @@
...
+    partial class ApplicationDbContextModelSnapshot : ModelSnapshot
     {
-        protected override void BuildTargetModel(ModelBuilder modelBuilder)
+        protected override void BuildModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
 #if (UseLocalDB)
-                .HasAnnotation("ProductVersion", "2.1.0-preview1")
+                .HasAnnotation("ProductVersion", "2.1.0-preview1-28193")
```
Eilon: @ajcvickers / @bricelam - just checking, the version here is OK? I know we've discussed this a billion times but I can never remember, sorry!

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
```diff
@@ -241,7 +240,8 @@ protected override void BuildTargetModel(ModelBuilder modelBuilder)
                         .HasForeignKey("UserId")
                         .OnDelete(DeleteBehavior.Cascade);
                 });
+
 #pragma warning restore 612, 618
-        }
+         }
```
Eilon: How did this extra space get here? The indentation is now off. Is this a bug in the migration generation logic itself?

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlServer/ApplicationDbContextModelSnapshot.cs
```diff
@@ -241,7 +240,8 @@ protected override void BuildTargetModel(ModelBuilder modelBuilder)
                         .HasForeignKey("UserId")
                         .OnDelete(DeleteBehavior.Cascade);
                 });
+
 #pragma warning restore 612, 618
-        }
+         }
```
Eilon: See earlier comment.

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
HaoK: These current files are still the result of the previous hand merge/edits.  The next PR will add the batch files that can be used to generate these for check in purposes, and I'll regenerate the migrations with the new batch files.

Eilon: Ah OK so I'll ignore for now. Thanx!

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
```diff
@@ -78,6 +78,24 @@
           "exclude": [
             "Extensions/AzureAdB2C*.cs"
           ]
+        },
+        {
```
seancpeters: There appears to be a conflict between these new modifiers and this existing modifier:
```
       {
          "condition": "(!IndividualLocalAuth)",
          "exclude": [
            "Data/**",
            "Areas/**"
          ]
        },
```
If a user chooses auth = IndividualB2C, then that makes IndividualLocalAuth = false, causing the above condition to be true, which excludes the Data/ directory. 

These new modifiers, conditioned on UseLocalDB, want to include some of the content under the Data/ directory, but it's been excluded.

The existing modifier which excludes the Data/ dir might need to be revisited with this in mind.

seancpeters: Looking at the rest of this review, I noticed that for StarterWeb CSharp, the modifier to exclude Data/ & Areas/ is conditioned on "(!IndividualAuth)", which might be the desired change here.

HaoK: Well it shouldn't matter as the rename won't do anything if the top level Data was all excluded right?  But seems harmless to add an additional IndividualLocalAuth check for these new conditions...

seancpeters: Correct, the rename won't do anything if the Data folder is already excluded. 

But it seems to be the case that with auth = IndividualB2C, we'd want to include the appropriate migration files - based on the UseLocalDB value. If this is incorrect, then it's probably fine as is.

HaoK: IndividualB2C = true means IndividualLocalAuth = false...  so the migrations should and are excluded.  I'm happy to add the IndividualLocalAuth check if you think that makes it clearer in the long run...  its not exactly obvious how all this stuff plays together at all, so even redundant checks have some value for clarity in the future

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Data/SqlLite/ApplicationDbContextModelSnapshot.cs
ajcvickers: @Eilon It doesn't need to match anything so this is fine for preview, but we should update it before RTM ships--I filed https://github.com/aspnet/templating/issues/353 for this and assigned it to @HaoK

HaoK:
Updated cleaning up the ApplicationDbContextSnapshot if defs, and added extra IndividualLocalAuth checks to the conditions, look good @seancpeters ?  The files on disk are the same either way, so functionally I don't think it makes a difference

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
seancpeters: I don't feel too strongly either way about adding the additional check.

The reason brought this up originally is because the description on the UseLocalDB parameter indicates that it is applicable for both Individual & IndividualB2C auth types. With that in mind, I assumed that migrations would be desired for B2C.

HaoK: These are IndividualAuth specific migrations so that's why

Issue #339, opened 2018-03-06
Update ValuesController in Web API project template to use ActionResult<T>


DamianEdwards:
The `ValuesController` in the Web API project template should be updated to return `ActionResult<T>` from the two object returning actions it has:

``` c#
// GET api/values
[HttpGet]
public ActionResult<IEnumerable<string>> Get()
{
    return new string[] { "value1", "value2" };
}

// GET api/values/5
[HttpGet("{id}")]
public ActionResult<string> Get(int id)
{
    return "value";
}
```

@Eilon @danroth27

mkArtakMSFT:
@jbagga, assigning this to you so you can keep it on your radar and merge when ready.

PR #351, opened 2018-03-10
Convert ValuesController return types to ActionResult<T>

scottaddie:
Fixes https://github.com/aspnet/templating/issues/339

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
@@ -18,14 +18,14 @@ namespace Company.WebApplication1.Controllers
     {
         // GET api/values
         [HttpGet]
-        public IEnumerable<string> Get()
+        public ActionResult<IEnumerable<string>> Get()
         {
             return new string[] { "value1", "value2" };
         }
 
         // GET api/values/5
         [HttpGet("{id}")]
-        public string Get(int id)
+        public ActionResult<string> Get(int id)
         {
             return "value";
         }
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
@@ -13,11 +13,13 @@ type ValuesController () =
 
     [<HttpGet>]
     member this.Get() =
-        [|"value1"; "value2"|]
+        let values = [|"value1"; "value2"|]
+        ActionResult<string[]>(values)
 
     [<HttpGet("{id}")>]
     member this.Get(id:int) =
-        "value"
+        let value = "value"
+        ActionResult<string>(value)
 
     [<HttpPost>]
     member this.Post([<FromBody>] value:string) =


Eilon:
Adding a few reviewers who might be interested: @jbagga @DamianEdwards @rynowak @pranavkm .

Eilon:
@jbagga - putting on your plate for final review + merge.

Eilon:
@cartermp - what would be the syntax for this, if any, in F#? Here's the F# ValuesController:

https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs#L14-L20

cartermp:
Was this change made more as a consistency thing, or is there a different reason?

I don't think there should be any change required for F# here.

Reason: The return types inferred by the compiler will still be `string []` and `string`, but that gets correctly dealt with by ASP.NET anyways. Working directly with `ActionResult<'T>` in F# is a bit cumbersome because we don't do implicit conversions like C# does. That means that something like this:

```cs
public async Task<ActionResult> Register(RegisterViewModel model)
{
    if (someCondition)
    {
        return RedirectToAction(foo, "bar", "wheee");
    }
    
    return SomethingElse();
}
```

Is a bit more annoying to encode because you must ensure that `SomethingElse()` either have the same return type as `RedirectoToAction` or cast to the appropriate base type or interface. F# folks are also far less likely to build fully-fledged MVC-style apps which utilize this heavily, so I wouldn't say it's important for F# unless there's a technical reason to make everything be `ActionResult<'T>` based.

Eilon:
@cartermp having that explicit return type is optional, but it helps MVC do a few things:

1. It can provide better metadata about the structure of the response (e.g. what allowed response types there are)
2. It allows the developer to either return the data type (e.g. `string[]`), or to return an HTTP status code such as a 404 Not Found

So it's very much a recommended pattern now, and I think even F# users would benefit from it (though the syntax might be less than ideal in F#).

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
```diff
@@ -18,14 +18,14 @@ public class ValuesController : ControllerBase
     {
         // GET api/values
         [HttpGet]
-        public IEnumerable<string> Get()
+        public ActionResult<IEnumerable<string>> Get()
```
davidfowl: This one is kinda meh.

Eilon: @davidfowl care to elaborate?

davidfowl: Nested generics don't look great. It's also no async (I assume on purpose).

cartermp:
@Eilon I see. In that case, the F# syntax update would be simple:

```fsharp

    [<HttpGet>]
    member this.Get() =
        let values = [|"value1"; "value2"|]
        ActionResult<string []>(values)

    [<HttpGet("{id}")>]
    member this.Get(id:int) =
        let value = "value"
        ActionResult<string>(value)
```

Those could be one line for bikeshedding purposes, but I think it's better to just have it like this.

I also noticed that I get "Connection is not secure" errors when testing with dotnet run:

screen shot 2018-03-12 at 10 12 14 am
 
With these tools:

```
.NET Command Line Tools (2.1.300-preview1-008174)

Product Information:
 Version:            2.1.300-preview1-008174
 Commit SHA-1 hash:  b8df89a54f

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.13
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.1.300-preview1-008174/

Microsoft .NET Core Shared Framework Host

  Version  : 2.1.0-preview1-26216-03
  Build    : f2c3216183d20416568a4bbf5bb7d153e826f153
```

Could that be issue in the recent updates to use HTTPS? The template is updated and the code looks like this:

```fsharp
    member this.Configure(app: IApplicationBuilder, env: IHostingEnvironment) =
        if (env.IsDevelopment()) then
            app.UseDeveloperExceptionPage() |> ignore
        else
            app.UseHsts() |> ignore

        app.UseHttpsRedirection() |> ignore
        app.UseMvc() |> ignore
```

Which looks correct based on what the [C# template is doing](https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Startup.cs#L48).

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
Eilon: Well of course this method can't  be async, it's trivially sync...

And the developer is expected to modify this method to their suiting, so they can change the return type and async-ness as needed once they fill in their code.

Eilon:
@cartermp thanks for the updated code, I think we should use that. @scottaddie - can you update the F# template code for consistency with C#?

@cartermp - the error you're seeing is expected if you haven't enabled the HTTPS dev cert. When you run "dotnet" for the first time it should have given you instructions to enable the dev cert. You need to run `dotnet dev-certs https --trust`, and we have more info on this blog post: https://blogs.msdn.microsoft.com/webdev/2018/02/27/asp-net-core-2-1-https-improvements/

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
```diff
@@ -13,11 +13,13 @@ type ValuesController () =
 
     [<HttpGet>]
     member this.Get() =
-        [|"value1"; "value2"|]
+        let values = [|"value1"; "value2"|]
+        ActionResult<string[]>(values)
```
jbagga: @cartermp If someone wanted to return Found or some other ActionResult will they replace `ActionResult` here? I am not sure I understand how this guides anyone using the template to use the recommended pattern @Eilon mentioned [here](https://github.com/aspnet/templating/pull/351#issuecomment-372380792)

cartermp: Hmm, I've only ever done ASP.NET MVC without explicit ActionResults in F#, which makes things a lot easier. Doing it explicitly like this makes it awkward, since you'll need to have all return paths casted to a base class or interface (not sure which would be best for MVC).

Making this template a pathway to learning that would be a challenge due to the lack of implicit conversions. I don't think it's particularly harmful to have these explicitly right now, but I think a lot of F# developers will just go back to returning the data as-is.

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
davidfowl:
```C#
public Task<ActionResult<IEnumerable<string>>> Get() => Task.FromResult(new [] { "zomg" });
```

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-FSharp/Controllers/ValuesController.fs
Eilon: @jbagga I think that if you start from this C#:

```c#
if (blah)
{
    return new string[] { "one", "two" };
}
else
{
    return NotFoundResult(); // helper factory method
}
```

Then after the explicit conversions, this code effectively ends up being:

```c#
if (blah)
{
    return new ActionResult<string[]>(new string[] { "one", "two" });
}
else
{
    return new ActionResult<string[]>(NotFoundResult()); // helper factory method
}
```

And in C# because the declared return type would be `ActionResult<string[]>`, the MVC metadata system can deduce many things about the result.

So in F#, though I don't know what the code would exactly look like, I hope you can do something like this totally made up code:

```f#
        blah ??
            if true ActionResult<string[]>(values)
            if false ActionResult<string[]>(this.NotFoundResult)
```

And then if the F# compiler figures out that the declared return type is `ActionResult<string[]>`, then I think everything will work.

cartermp: @Eilon yep!

```fsharp
    [<HttpGet("{id}")>]
    member this.Get(id:int) =
        if id > 100 then
            ActionResult<string>("Over 100!")
        else
            ActionResult<string>(NotFoundResult())
```

And this works as you'd expect - I get a 404 not found if `id` is 100 or less. I didn't actually know that you can pass `NotFoundResult` into an `ActionResult`. Fun times.

jbagga: Ok then this looks good! Thanks @Eilon and @cartermp. Approving the PR now

src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/Controllers/ValuesController.cs
Eilon: Not going to feed the TROLLS. You troll.

jbagga:
Merging now. Thanks @scottaddie

scottaddie:
Happy to help! Thanks to everyone for reviewing.

Issue #346, opened 2018-03-08
LaunchSettings profile Company.WebApplication1 is created insted of  profile <AppName>


btorrild:
Hello, when applying a new angular project template using 2.1 preview1-final the default non IIS launch profile is called Company.WebApplication1.

Should that not be the name of the console app?

Eilon:
@btorrild hmm that certainly doesn't look right, we might be missing a substitution parameter somewhere. Can you show a screenshot or share more info about exactly where you're seeing this? Thanks!

btorrild:
Does this help. 
```
PS>dotnet --version
2.1.300-preview1-008174

PS>dotnet new angular --name test
The template "ASP.NET Core with Angular" was created successfully.
The usual stuff here...

PS>cat .\test\Properties\launchSettings.json
{
  "iisSettings": {
    "windowsAuthentication": false,
    "anonymousAuthentication": true,
    "iisExpress": {
      "applicationUrl": "http://localhost:45041",
      "sslPort": 44385
    }
  },
  "profiles": {
    "IIS Express": {
      "commandName": "IISExpress",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "ASPNETCORE_HTTPS_PORT": "44385"
      }
    },
    "Company.WebApplication1": {
      "commandName": "Project",
      "launchBrowser": true,
      "environmentVariables": {
        "ASPNETCORE_ENVIRONMENT": "Development",
        "ASPNETCORE_URLS": "https://localhost:5001;http://localhost:5000"
      }
    }
  }
}

```
I tried some preview templates for 2.0, could that be an issue

Eilon:
@btorrild thanks for the details, that is very helpful. This indeed looks like a bug so we'll take a look at it.

Eilon:
Note: We'll need to investigate whether we should consider patching this or if it's a regression in 2.1.

mkArtakMSFT:
@jbagga can you please look into this? Thanks! Also, assign a cost, please.

jbagga:
launchSettings.json was added in 2.1.0-preview1 https://github.com/aspnet/templating/commit/4d2e642b94ad6c2e152e09005c33bac7543bcf04
Sending a PR for dev

PR #354, opened 2018-03-12
Match launchSettings variable to sourceName in template.json

jbagga:
Addresses #346 

cc @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
+++ b/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
@@ -16,7 +16,7 @@
         "ASPNETCORE_HTTPS_PORT": "44300"
       }
     },
-    "AureliaSpa": {
+    "Company.WebApplication1": {
       "commandName": "Project",
       "launchBrowser": true,
       "environmentVariables": {
--- a/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
+++ b/src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
@@ -16,7 +16,7 @@
         "ASPNETCORE_HTTPS_PORT": "44300"
       }
     },
-    "Company.WebApplication1": {
+    "AureliaSpa": {
       "commandName": "Project",
       "launchBrowser": true,
       "environmentVariables": {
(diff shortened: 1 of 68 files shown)


jbagga:
It looks like the launchSettings.json did not match the sourceName defined in template.json before.

src/Microsoft.AspNetCore.SpaTemplates/content/Aurelia-CSharp/Properties/launchSettings.json
```diff
@@ -16,7 +16,7 @@
         "ASPNETCORE_HTTPS_PORT": "44300"
       }
     },
-    "Company.WebApplication1": {
+    "AureliaSpa": {
```
seancpeters: To help deal with when the user inputs a name for a template that is not valid in some contexts (namespace for example) template processing attempts to create valid variations on the input name, so that the generated template content will build. But to do this, the sourceName must be in a format that will be properly dealt with when applying the variations. It's for this reason that we generally use a name like "Company.WebApplication1".

So I'd recommend making the change in the opposite direction, that is, leave the source name as Company.WebApplication1 in the template.json, and change "AureliaSpa" to "Company.WebApplication1" in the template content files. Otherwise when users enter names that aren't valid symbols in the context where they get replaced, the generated template won't build.

(Same thing for all the other templates with similar changes in this PR).

jbagga:
@seancpeters Updated. 

cc FYI @SteveSandersonMS to ensure no inadvertent changes were made to any spa templates

seancpeters:
Looks good to me. For testing, I'd recommend trying to create the varios templates with names that would be invalid namespaces / class names, etc. Things like "C.1", "C!", "foo-bar", and see if the generated content builds / has the expected replacements.

jbagga:
After changing the name of the generated csproj, the generated content builds fine. "Vue.1!" is replaced by "Vue._1_" Is that expected? I was expecting Vue.1_ in launchSettings
Others like "Angular.!" are replaced by "Angular._" in launchSettings

seancpeters:
The reason that "Vue.1!" gets replaced by `Vue._1_` is because in C# namespaces, the character immediately following a "." cannot be a digit. So when there is a digit immediately after a ".", we insert a "_" before the digit to make the namespace name valid. The "!" is invalid anywhere in a namespace, and all strictly invalid characters also get replaced with "_". 

"Angular.!" becomes "Angular._" because the only transform needed is to replace the invalid "!" with a "_".

jbagga:
@Eilon launchSettings were added in 2.1.0-preview1 so can this just be for dev? Don't think we need a patch?

PR #356, opened 2018-03-14
Made wwwroot 'copyOnly' to avoid modifications in it's files

### LINKED ISSUE #331: Files under wwwroot getting incorrect replacements
The templates RazorPagesWeb-CSharp, StarterWeb-CSharp, and StarterWeb-FSharp all define the "HttpPortReplacer" symbol in their template.json files, setup to replace "8080". But the value "8080" appears in various files under the wwwroot directory in each of these templates, and thus have their instances of 8080 replaced by the value assigned to HttpPortReplacer.

The easiest way to deal with this is probably to change the template.json's for those templates so that everything under the wwwroot directory is copyOnly - assuming that no template processing should be performed on those files. Changing to copyOnly also has the side-benefit that the template invocation will be a bit faster.

If any template processing other than copying the files in the wwwroot directory needs to take place, I can help figure out the best way to deal with this.

kichalla:
[Fixes #331] Files under wwwroot getting incorrect replacements

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
@@ -26,6 +26,9 @@
   ],
   "sources": [
     {
+      "copyOnly": [
+        "wwwroot/**"
+      ],
       "modifiers": [
         {
           "condition": "(TargetFrameworkOverride == '')",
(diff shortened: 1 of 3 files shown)


jbagga:
@kichalla Were you able to create these templates and ensure they work as expected? And also if the files in this directory are ignored when trying to replace 8080 (as in the original bug)?

kichalla:
> Were you able to create these templates and ensure they work as expected? And also if the files in this directory are ignored when trying to replace 8080 (as in the original bug)?

I initially verified using manually (checked in original wwwroot to git and then `dotnet new` to the folder to see if there are any changes to wwwroot), but i have now done an end-to-end test of the application with my changes and they work fine as expected (i.e wwwroot content is not modified after my changes)

kichalla:
https://github.com/aspnet/templating/commit/fa8fcc9a628153f17c5db2d3275179b07548bf77
https://github.com/aspnet/templating/commit/8f2c029a95366e5037dcd1812a6c9a7e776cfd25

PR #360, opened 2018-03-15
Removing launchSettings.json from project templates

artiomchi:
I noticed that the new templates used in ASP.NET Core 2.1 have the `launchSettings.json` files present. The templates in the previous version of the framework didn't have these files.

Seems like the files contain the default settings for the app anyway, so they're not providing much value in the first place.

The main issue though is the fact that their presence will prevent overriding settings using environment variables.

For example, consider this:

```
docker run -p 3001:5001 -p 3000:5000 -it --rm --name dotnetwebtest -e ASPNETCORE_URLS=https://*:5001 microsoft/dotnet:2.1-sdk-alpine

# dotnet new mvc -n testapp
# cd testapp
# dotnet run
Using launch settings from /testapp/Properties/launchSettings.json...
[...]
Hosting environment: Development
Content root path: /testapp
Now listening on: https://localhost:5001
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
```

The app ignores the environment variable I passed because it's overridden in launchSettings.json

I suggest reverting back to previous behaviour by removing the launch settings files, to make the templates create an app that behaves as described in the docs.


kasper3:
Another approach is https://github.com/aspnet/Hosting/issues/1326, to be able to override all those options via command line arguments while keeping the launchSettings.json (which has its significance).
CC @davidfowl

artiomchi:
Hmm.. I don't think that issue is relevant here, tbh. This PR tries to fix an issue with the templates that was added in 2.1, and bring them back to 2.0 behaviour

Eilon:
@DamianEdwards / @danroth27 / @javiercn - thoughts on this?

danroth27:
The default environment is Production, not Development. We want the environment to be Development when doing `dotnet run`.

We could remove setting ASPNETCORE_URLS from `launchSettings.json`. We only had it there as a convenience so that folks can discover a way to set it. But maybe it causes more harm than good.

DamianEdwards:
The behavior was changed intentionally, namely to get the environment set by default in CLI scenarios, and as part of the new HTTPS by default work. But as Dan says, we can likely remove the URLS var without issue.

javiercn:
dotnet run is a development command hence it runs with development settings. This makes the experience of running the app from command line with dotnet run and running the app from visual studio similar and its a deliberate choice.

For docker we've moved to a multi-step build process, using an image to build the app and another image to run the built app, inside docker you should always run your app with dotnet <<path-to-dll>> -- <<args>>

@danroth27 I would be ok with tooling respecting existing environment variables over those defined in/by launchSettings.json but that would be a tooling change.

Eilon:
@danroth27 / @DamianEdwards / @javiercn - can you provide more clarity on exactly what change(s) you propose would work for us?

javiercn:
We can't revert anything on this file, only how tooling interprets it and what it gives precedence to.
ASPNETCORE_URLS will go away in preview2 in favor of applicationUrl, but that won't change anything as the meaning tooling gives applicationUrl is to setup ASPNETCORE_URLS and that would override anything defined in your environment.

The only thing we could consider doing is to change tooling to prefer/preserve things already in your environment.

You should not run your app inside a container with `dotnet run` but with `dotnet <<path-to-dll>>`. dotnet run is intended for development and to use launchSettings.json

artiomchi:
@danroth27 @DamianEdwards @javiercn I see, this makes perfect sense.

Removing the environment variable won't break the https by default work, since it has the default value anyway. Also you can see which urls the app runs on in the output of dotnet run.

I can certainly see the point of making this feature more discoverable though. So how about this option then? Json doesn't support commenting options, but nothing stops us from naming the property `"//ASPNETCORE_URLS"` in json. It's discoverable, and easy to see as a commented field. Does that look better?

artiomchi:
Having thought more on this and reading the comments above, I realised that this wasn't really a necessary change in the first place. Even in the edge cases when someone would want to run the app using dotnet run vs running a deployed version (unit testing? during development using containers?) they could just as easily configure their launchSettings.json and commit that to their repo, so this change is indeed unnecessary.

Now I feel silly after wasting everyone's time on this 😅

javiercn:
@artiomchi Don't worry about it. We are here to help :)

Issue #340, opened 2018-03-06
Move all partial & layout views in Razor Pages project template to Shared folder


DamianEdwards:
In 2.1 we added support for the *Shared* folder under the configured Pages root. In the project templates however we have a mix of shared assets (partials, layouts, etc.) in the pages root itself, and the shared folder.

We should make this consistent and move all these assets into the new */Pages/Shared* folder in the project template.

@Eilon

mkArtakMSFT:
@jbagga , can you please handle this? Thanks!

PR #368, opened 2018-03-16
Move layout and partial script to Shared dir

jbagga:
Addresses #340


korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-preview2-15742
-commithash:21fbb0f2c3fe4a9216e2d59632b98cfd7d685962
+version:2.1.0-preview3-15744
```
pranavkm: Did this update when you did `build -update`? It shouldn't be picking up `preview3`.

pranavkm: Could you re-run this?

javiercn:
We should also move _LoginPartial if its not already there.

jbagga:
I removed the korebuild file @pranavkm 
@javiercn Added the template.json condition

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
```diff
@@ -60,6 +60,12 @@
             "Areas/**"
           ]
         },
+        {
+          "condition": "(IndividualB2CAuth)",
+          "rename": {
+            "Pages/_LoginPartial" : "Pages/Shared/_LoginPartial"
```
Eilon: File extensions aren't needed here?

jbagga: Not needed but I think nice to have. I'll add them

Eilon: Yeah it *looks* wrong, even if you assure me it's right 😄

jbagga:
@Eilon Updated

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
```diff
@@ -60,6 +60,12 @@
             "Areas/**"
           ]
         },
+        {
+          "condition": "(IndividualB2CAuth)",
+          "rename": {
+            "Pages/_LoginPartial.cshtml" : "Pages/Shared/_LoginPartial.cshtml"
```
javiercn: I would have them living side by side in the `Pages/Shared` folder, name each of them explicitly `_LoginPartial.Identity.cshtml` and `_LoginPartial.OrgAuth.cshtml` and just make the rename be from `_LoginPartial.Identity.cshtml -> _LoginPartial.cshtml` and from `_LoginPartial.OrgAuth.cshtml -> _LoginPartial.cshtml`

The reason for it is that is not obvious what case each _LoginPartial file is for when you are editing them

Eilon: That sounds like a good idea to me.

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json
```diff
@@ -60,6 +60,12 @@
             "Areas/**"
           ]
         },
+        {
+          "condition": "(IndividualB2CAuth)",
```
javiercn: This condition is wrong, it should include Individual/MultiOrg

jbagga:
I have modified existing conditions to rename and exclude appropriately @javiercn

jbagga:
ping @pranavkm @javiercn Updated

Issue #276, opened 2018-02-01
Observed some BOM/non ASCII characters in appsettings.json in .net core web api project


Pilchie:
_From @shinkarmv on January 22, 2018 16:42_

While working with appsettings transformations using python script in Docker file.
I have observed that some BOM/non ASCII characters in the starting of appsettings.json

Please find screenshot below for reference
![image](https://user-images.githubusercontent.com/31374791/35232511-a587348a-ffc1-11e7-9f82-287cded2edfe.png)

These characters came by default with project template of WEB API in Visual Studio 2017 .Net Core.

Regards,
Mukesh

_Copied from original issue: dotnet/project-system#3171_

Pilchie:
Likely a UTF-8 BOM?

Pilchie:
@BillHiebert ?

Pilchie:
_From @shinkarmv on January 22, 2018 16:54_

Yes @Pilchie.  These are BOM characters.

Pilchie:
_From @shinkarmv on January 22, 2018 17:49_

I have cross verified the unit test at SchemaStore repository and observed no BOM character in it.

Please see below screenshot for reference.

![image](https://user-images.githubusercontent.com/31374791/35235561-86c41afa-ffca-11e7-85c5-fca50eae9deb.png)

Regards,
Mukesh

Pilchie:
@jmarolf @mlorbetske - do you know where this comes from?

Pilchie:
_From @jmarolf on January 23, 2018 21:11_

@mlorbetske I believe these come from the WebToolsExtensions repo

Pilchie:
_From @shinkarmv on January 26, 2018 8:11_

Hi @Pilchie ,

I did some analysis on the aforementioned issue and finally got the Project Template for WEB API for C# .net core app.

Repository - https://github.com/aspnet/templating

Template link - https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/WebApi-CSharp/appsettings.json

I have validated appsettings.json file and got some BOM characters at start of file.
Please find the screenshot below for reference,

![image](https://user-images.githubusercontent.com/31374791/35430494-821a07ea-029e-11e8-83bc-c470e5ec2d7a.png)
 
cc: @mlorbetske, @jmarolf

Regards,
Mukesh

BillHiebert:
@Pilchie  Yes these come from the template. Is there a concern about the characters being there? @phenning

[...]

PR #369, opened 2018-03-17
Remove BOM from json files

jbagga:
Addresses #276 

1. Removed BOM from json files. 
2. Added a test to check if BOM present

The change does not show the BOM removed but I ran the test before and after and BOM has been removed from the below files.

### DIFF
--- a/test/Templates.Test/ByteOrderMarkTest.cs
+++ b/test/Templates.Test/ByteOrderMarkTest.cs
@@ -17,15 +17,16 @@ namespace Templates.Test
         }
 
         [Theory]
-        [InlineData("\\Microsoft.AspNetCore.SpaTemplates\\content")]
-        [InlineData("\\Microsoft.DotNet.Web.ProjectTemplates\\content")]
-        [InlineData("\\Microsoft.DotNet.Web.Spa.ProjectTemplates\\content")]
+        [InlineData(@"\Microsoft.AspNetCore.SpaTemplates\content")]
+        [InlineData(@"\Microsoft.DotNet.Web.ProjectTemplates\content")]
+        [InlineData(@"\Microsoft.DotNet.Web.Spa.ProjectTemplates\content")]
         public void CheckForByteOrderMarkSpaTemplates(string path)
         {
             var currentDirectory = Directory.GetCurrentDirectory();
             var srcDirectory = Path.GetFullPath(Path.Combine(currentDirectory, @"..\..\..\..\..\src"));
             var directories = Directory.GetDirectories(srcDirectory + path, "*Sharp");
-            var checkFiles = false;
+
+            var filesWithBOMCharactersPresent = false;
             foreach (var directory in directories)
             {
                 var files = Directory.GetFiles(directory, "*.json");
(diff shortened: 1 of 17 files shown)


test/Templates.Test/ByteOrderMarkTest.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly ITestOutputHelper _output;
+
+        public ByteOrderMarkTest(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Theory]
+        [InlineData("\\Microsoft.AspNetCore.SpaTemplates\\content")]
```
Eilon: Might want to use `@"..."` strings here too.

test/Templates.Test/ByteOrderMarkTest.cs
```diff
@@ -0,0 +1,58 @@
...
+        [InlineData("\\Microsoft.DotNet.Web.ProjectTemplates\\content")]
+        [InlineData("\\Microsoft.DotNet.Web.Spa.ProjectTemplates\\content")]
+        public void CheckForByteOrderMarkSpaTemplates(string path)
+        {
+            var currentDirectory = Directory.GetCurrentDirectory();
+            var srcDirectory = Path.GetFullPath(Path.Combine(currentDirectory, @"..\..\..\..\..\src"));
+            var directories = Directory.GetDirectories(srcDirectory + path, "*Sharp");
+            var checkFiles = false;
```
Eilon: Not sure I understand the semantics of this variable. Can we find a better name for it?

test/Templates.Test/ByteOrderMarkTest.cs
```diff
@@ -0,0 +1,58 @@
...
+            foreach (var directory in directories)
+            {
+                var files = Directory.GetFiles(directory, "*.json");
+                foreach (var file in files)
+                {
+                    var filePath = Path.GetFullPath(file);
+                    var fileStream = new FileStream(filePath, FileMode.Open);
+
+                    var bits = new byte[3];
```
Eilon: Bytes, not bits.

test/Templates.Test/ByteOrderMarkTest.cs
jbagga: Basically a flag that says the test output should be checked for files that have BOM characters.

Eilon: Isn't it saying the *result* of the test (success/failure), as opposed to whether something should be checked?

test/Templates.Test/ByteOrderMarkTest.cs
```diff
@@ -0,0 +1,59 @@
...
+        [InlineData(@"\Microsoft.AspNetCore.SpaTemplates\content")]
+        [InlineData(@"\Microsoft.DotNet.Web.ProjectTemplates\content")]
+        [InlineData(@"\Microsoft.DotNet.Web.Spa.ProjectTemplates\content")]
+        public void CheckForByteOrderMarkSpaTemplates(string path)
+        {
+            var currentDirectory = Directory.GetCurrentDirectory();
+            var srcDirectory = Path.GetFullPath(Path.Combine(currentDirectory, @"..\..\..\..\..\src"));
+            var directories = Directory.GetDirectories(srcDirectory + path, "*Sharp");
+
+            var filesWithBOMCharactersPresent = false;
```
jbagga: Is this any better? @Eilon

Eilon: Ah, new name is better!

jbagga:
Yes I have confirmed that all the files in this PR were reported and the test failed. It now passes with the BOM removed

PR #374, opened 2018-03-19
Add HostFiltering middleware to templates

Tratcher:
This adds the new HostFiltering middleware to every template except Empty and F#. 

~~TODO: Update the dependencies once the 2.1 branch [publishes some](https://dotnet.myget.org/feed/aspnetcore-release/package/nuget/Internal.AspNetCore.Universe.Lineup).~~
~~TODO: Disable the new middleware by default~~
~~TODO: File an issue to port the final version of this to the F# tempalates~~ https://github.com/aspnet/templating/issues/392

@DamianEdwards please look a the code in Startup. It's enabled right now so I can verify the dependencies. The proposal was to disable it by commenting it out. It may be cleaner to disable it by adding this config `"AllowedHosts": [ "*" ]`.

### DIFF
--- a/Templating.sln
+++ b/Templating.sln
@@ -14,6 +14,13 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Web.Client
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.DotNet.Web.ProjectTemplates", "src\Microsoft.DotNet.Web.ProjectTemplates\Microsoft.DotNet.Web.ProjectTemplates.csproj", "{260EBA09-DEF5-429C-99BF-90CA1456A576}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{F34CD7D7-1E8F-41FA-AB4D-D86BFD3AC09B}"
+	ProjectSection(SolutionItems) = preProject
+		build\dependencies.props = build\dependencies.props
+		build\repo.props = build\repo.props
+		build\sources.props = build\sources.props
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/appsettings.json
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/appsettings.json
@@ -35,5 +35,6 @@
     "LogLevel": {
       "Default": "Warning"
     }
-  }
+  },
+  "AllowedHosts": [ "*" ]
 }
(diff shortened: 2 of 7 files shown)


Eilon:
> TODO: Disable the new middleware by default

I don't really like the idea of the templates having commented-out code. Ultimately that's up to @DamianEdwards , though.

Also, you can rebase your branch because @kichalla just deleted some of the SPA templates that you're updating, so those changes are of course not needed anymore.

Tratcher:
@DamianEdwards ready for review.

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -55,6 +55,11 @@ public void ConfigureServices(IServiceCollection services)
                 options.MinimumSameSitePolicy = SameSiteMode.None;
             });
 
+            services.AddHostFiltering(options =>
+            {
+                options.AllowedHosts = Configuration.GetSection("AllowedHosts").Get<List<string>>();
```
DamianEdwards: Will `Bind` not work here? Could we do an overload of `AddHostFiltering` that takes the configuration section directly?

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -104,6 +109,9 @@ public void ConfigureServices(IServiceCollection services)
         // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
         public void Configure(IApplicationBuilder app, IHostingEnvironment env)
         {
+            // For use with publicly exposed servers that are not restricting the Host header.
```
DamianEdwards: I'm actually not sure what this comment is trying to say.

DamianEdwards:
I don't mind the idea of adding default configuration instead as per @Tratcher's suggestion, as to avoid the commented out lines. I hadn't thought about the services line too so having two lines in different methods commented out is especially urk.

DamianEdwards:
Remind me again why we decided this needed to be in the templates at all, given it's going to be disabled?

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
Tratcher: e.g. Kestrel on the edge.

Tratcher:
Discoverability?

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
Tratcher: Updated to use Bind

Tratcher:
Updated with configuration binding and a default config that disables the middleware with a wildcard.

Tratcher:
Offline notes: Damian doesn't want the extra lines added in Startup, he'd rather add this to CreateDefaultWebHostBuilder. If there's no config then disable the middleware. He'll confirm with fowler.

Also, use a semicolon separated list to mirror URLs. It's easier to specify from command line, env vars, etc..

PR #386, opened 2018-03-20
Removing Microsoft.AspNetCore.SpaTemplates project

### LINKED ISSUE #176: KnockoutTemplate_Works_NetCore test flaky on VS
`KnockoutTemplate_Works_NetCore` test fails with the following error when run as part of a test run in VS

Expected: Weather forecast
Actual:   Counter

If the test is run by itself, it passes.

Possible race condition? https://github.com/aspnet/templating/blob/dev/test/Templates.Test/SpaTemplateTest/SpaTemplateTestBase.cs#L72 and https://github.com/aspnet/templating/blob/dev/test/Templates.Test/SpaTemplateTest/SpaTemplateTestBase.cs#L82.

cc @Eilon @SteveSandersonMS

kichalla:
[Fixes #176] KnockoutTemplate_Works_NetCore test flaky on VS


kichalla:
Removing the project and tests as per the announcement here: https://github.com/aspnet/Announcements/issues/289

Issue #328, opened 2018-02-17
Set the applicationUrl property instead of using the ASPNETCORE_URLS env var


danroth27:
In VS tooling the application URLs are specified using the applicationUrl property and is preferred to using the ASPNETCORE_URLS env var.

danroth27:
@BillHiebert @javiercn

javiercn:
@danroth27 This is a temporary workaround due to the lack of tooling support in preview1.

PR #390, opened 2018-03-21
Use applicationUrl instead of ASPNETCORE_URLS

jbagga:
Addresses #328


src/Microsoft.DotNet.Web.ProjectTemplates/content/EmptyWeb-CSharp/Properties/launchSettings.json
```diff
@@ -19,9 +19,9 @@
     "Company.WebApplication1": {
       "commandName": "Project",
       "launchBrowser": true,
+      "applicationUrl": "https://localhost:5001;http://localhost:5000",
```
javiercn: Can you also remove this? 
https://github.com/aspnet/templating/pull/390/files#diff-bd9b64bdb7d7d33da368a9efbf4d1cfaL16

It should not be necessary anymore (try it out), I think @danroth27 failed to mention it in the bug.

jbagga:
@javiercn Updated

Issue #394, opened 2018-03-22
Remove the BrowserLink package and middleware line references from the project templates


DamianEdwards:
It's disabled by default in VS today and we don't want to have an extra package ref and code in the project for the limited utility BrowserLink provides right now. We'll revisit how to make enabling it easy in the future.

DamianEdwards:
@Eilon to triage & assign pls

Eilon:
@jbagga - all yours.

jbagga:
BrowserLink package is brought in and the middleware is used only when UseBrowserLink is set to true
https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/.template.config/template.json#L230

By default it is false and therefore not used. Are we removing the UseBrowserLink option? @DamianEdwards

Eilon:
Just to make sure it's documented: we decided to keep the option, but turn it off by default for VS scenarios.

PR #395, opened 2018-03-22
Set useBrowserLink to false for VS host file

jbagga:
See #394


jbagga:
I spoke to @Eilon about that. The decision was to turn it off by default for VS. For CLI the default is false already

Issue #396, opened 2018-03-22
Remove references to Microsoft.EntityFrameworkCore.Tools from the project templates where Microsoft.AspNetCore.App is referenced.


JunTaoLuo:
As titled. Microsoft.EntityFrameworkCore.Tools is a dependency of Microsoft.AspNetCore.App and the explicit reference to Microsoft.EntityFrameworkCore.Tools is redundant.

PR #397, opened 2018-03-23
Remove EF.Tools reference

JunTaoLuo:
#396


JunTaoLuo:
fyi @muratg @Eilon

jbagga:
Could you remove the package from dependencies.props?

JunTaoLuo:
@jbagga We can't do that, the package itself is still used during when TargetFrameworkOverride is set, for example for full framework. For example: https://github.com/aspnet/templating/blob/release/2.1/src/Microsoft.DotNet.Web.ProjectTemplates/StarterWeb-CSharp.csproj.in#L38

PR #407, opened 2018-03-26
Set DebugType=full for .NET Framework projects

### LINKED ISSUE #222: Specify DebugType = Full for desktop targeting projects < net471
We made a change to Mvc as part of 2.1.0-preview to use the app's DebugType setting to determine the view's pdb type. The SDK defaults this value to `portable` which does not correctly work for desktop applications targeting < net471. We'll workaround this in our templates by explicitly specifying `DebugType = Full` in our desktop templates when the tfm < net471

pranavkm:
Fixes #222


pranavkm:
@Eilon I haven't modified the EmptyWeb \ WebAPI templates, and F# templates. Let me know if any of those need to be modified.

pranavkm:
> I assume that's because we care only about templates where there's meaningful Razor codez? If so, that's fine.

Yup.

jbagga:
Looks good to me

PR #307, opened 2018-02-07
Update jQuery

jbagga:
This reverts commit 382a762a08d072b235cd230c0fd883b3871c5799.

Addresses #18

PR #430, opened 2018-04-02
Update AJAX CDN links

jbagga:


src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Pages/Shared/_Layout.cshtml
```diff
@@ -59,7 +59,7 @@
         <script src="~/js/site.js" asp-append-version="true"></script>
     </environment>
     <environment exclude="Development">
-        <script src="https://ajax.aspnetcdn.com/ajax/jquery/jquery-2.2.0.min.js"
+        <script src="https://ajax.aspnetcdn.com/ajax/jquery/jquery-3.3.1.min.js"
```
Eilon: Do we need to update the version of jQuery as well?

jbagga: We updated the fallback jQuery libraries already. https://github.com/aspnet/templating/pull/307

Eilon: Oh we actually made that the required minimum? I completely forgot. Looks good then.

Issue #429, opened 2018-04-02
Use dotnet-ef package for the tests


jbagga:
Right now we depend on the CLI to install dotnet-ef and use it in our tests.

They should instead be decoupled and use Korebuild to download and install "dotnet-ef" and consume in our tests.

cc @natemcmaster @mkArtakMSFT

jbagga:
Thank you @natemcmaster!

Issue #378, opened 2018-03-19
Re-enable Auth tests


jbagga:
Waiting for dotnet-ef to be in the CLI so it can be consumed from our Auth tests in templating.

See https://github.com/aspnet/templating/pull/370

cc @ryanbrandenburg FYI

PR #433, opened 2018-04-03
Use dotnet-ef package

jbagga:
Address #429 and #378

### DIFF
--- a/Templating.sln
+++ b/Templating.sln
@@ -28,6 +28,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{342602FC-9
 		test\TemplateTests.props.in = test\TemplateTests.props.in
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DotNetToolsInstaller", "DotNetToolsInstaller\DotNetToolsInstaller.csproj", "{89317EDE-A5B9-4688-A74A-DC9FBDA71438}"
+EndProject
+
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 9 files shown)


test/Templates.Test/Templates.Test.csproj
```diff
@@ -26,4 +25,11 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualStudioPackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
```
pranavkm: Can `Template.Test` reference `DotnetToolInstaller`? Running these tests require the repo to be built, but it would be nice to explicitly express the dependency

test/Templates.Test/Helpers/TemplateTestBase.cs
```diff
@@ -132,7 +132,13 @@ protected void RunDotNet(string arguments)
 
         protected void RunDotNetEfCreateMigration(string migrationName)
         {
-            var args = $"ef migrations add {migrationName}";
+            var assembly = typeof(TemplateTestBase).Assembly;
+
+            var DotNetEfFullPath = assembly.GetCustomAttributes<AssemblyMetadataAttribute>()
+                .FirstOrDefault(attribute => attribute.Key == "DotNetEfFullPath")
```
Eilon: Can probably call `.First()` instead of `.FirstOrDefault()` considering that if there's no item in the list, the next line will throw a less meaningful NullRef anyway.

test/Templates.Test/Helpers/TemplateTestBase.cs
```diff
@@ -132,7 +132,13 @@ protected void RunDotNet(string arguments)
 
         protected void RunDotNetEfCreateMigration(string migrationName)
         {
-            var args = $"ef migrations add {migrationName}";
+            var assembly = typeof(TemplateTestBase).Assembly;
+
+            var DotNetEfFullPath = assembly.GetCustomAttributes<AssemblyMetadataAttribute>()
```
Eilon: `dotNetEfFullPath`

test/Templates.Test/Helpers/TemplateTestBase.cs
```diff
@@ -132,7 +132,13 @@ protected void RunDotNet(string arguments)
...
+                .FirstOrDefault(attribute => attribute.Key == "DotNetEfFullPath")
+                .Value;
+
+            var args = $"{DotNetEfFullPath} migrations add {migrationName}";
```
natemcmaster: Add quotes around the path just in case this is running on a machine with a space in the folder name.

Templating.sln
```diff
@@ -28,6 +28,9 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{342602FC-9
 		test\TemplateTests.props.in = test\TemplateTests.props.in
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "DotNetToolsInstaller", "DotNetToolsInstaller\DotNetToolsInstaller.csproj", "{89317EDE-A5B9-4688-A74A-DC9FBDA71438}"
```
natemcmaster: Nit: maybe put this under the tests folder?

Eilon: Ah good point. Doesn't deserve its own top-level folder.

PR #434, opened 2018-04-03
Use dotnet-ef package (#433)

jbagga:
Addresses  #429 and #378


jbagga:
Resolving a merge conflict with deps.props. I'll check in after that https://github.com/aspnet/templating/pull/435. Thanks

Issue #455, opened 2018-04-06
WebApi launch settings regression, 404


Tratcher:
In the new 2.1-preview2 templates the WebApi project always launches to the application root "/" and 404's. In 2.0 it would launch to "/api/values" and return some json. See `"launchUrl": "api/values"` that has been removed from the launchSettings.json file.

@jbagga  @javiercn

jbagga:
FYI @mkArtakMSFT  

https://github.com/dotnet/templating/issues/1508 There is an issue with the `launchSettings.json` schema. But it isn't blocking.

Issue #445, opened 2018-04-04
[Preview2-Validation] ReturnUrl in _LoginPartial is incorrect


kichalla:
Found this when verifying https://github.com/aspnet/Release/issues/259

A project based on MVC template + IndivualAuth option generates the following content in `Views\Shared\_LoginPartial.cshtml`

```html
@using Microsoft.AspNetCore.Identity

@inject SignInManager<ApplicationUser> SignInManager
@inject UserManager<ApplicationUser> UserManager

@if (SignInManager.IsSignedIn(User))
{
    <form asp-area="Identity" asp-page="/Account/Logout" asp-route-returnUrl="@Url.Page("/Index", new { area = "" })" method="post" id="logoutForm" class="navbar-right">
        <ul class="nav navbar-nav navbar-right">
            <li>
                <a asp-area="Identity" asp-page="/Account/Manage/Index" title="Manage">Hello @UserManager.GetUserName(User)!</a>
            </li>
            <li>
                <button type="submit" class="btn btn-link navbar-btn navbar-link">Logout</button>
            </li>
        </ul>
    </form>
}
else
{
    <ul class="nav navbar-nav navbar-right">
        <li><a asp-area="Identity" asp-page="/Account/Register">Register</a></li>
        <li><a asp-area="Identity" asp-page="/Account/Login">Login</a></li>
    </ul>
}
```

Notice that the `asp-route-returnUrl` should have been `asp-route-returnUrl="@Url.Action("Index", "Home", new { area = "" })"` instead of what currently is. This is a regression caused by this commit: https://github.com/aspnet/templating/commit/a8246e2a9906c647b7c444f735152a9f95534f97#diff-5ef9df60b5f09bb055f023e579a4ff00L28

Issue #392, opened 2018-03-22
F# Add HostFiltering middleware to non-empty templates


Tratcher:
Port of https://github.com/aspnet/templating/pull/374 to F# templates

PR #463, opened 2018-04-10
WebAPI launchUrl; MVC _LoginPartial; F# AllowedHosts

jbagga:
Addresses 
#455 : add a `launchUrl` for Web API templates RESOLVED: ~(issue with `launchSettings` schema, throws a warning when set to "api/values". It has to be a uri https://github.com/dotnet/templating/issues/1508)~
#392 : `AllowedHosts` config for F# non-empty templates
#445 : correct the `asp-route-returnUrl` for `_LoginPartial` for MVC C# app

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -3,40 +3,40 @@
     <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
   </PropertyGroup>
   <PropertyGroup Label="Package Versions">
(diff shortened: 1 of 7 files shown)


jbagga:
Update: Once https://github.com/SchemaStore/schemastore/pull/430 is in, the launchUrl can be simplified to "api/values"

Tratcher:
Want to wait for 430 so you don't have to change it twice?

jbagga:
@Tratcher Updated

Issue #469, opened 2018-04-12
Remove "Bower" from images that appear in Web Templates


justcla:
The web template for ASP.NET Core MVC web app includes four banner images. One of the images is about "Packages" and lists Bower (along with NuGet, npm and Gulp).

Bower is no longer supported by Visual Studio, so we should update the image, or remove it from the templates.

![image](https://user-images.githubusercontent.com/17131343/38705439-6aa1d326-3e5e-11e8-95e2-22c9d11bfa29.png)

justcla:
Note, please check all other web templates for use of this image.

DamianEdwards:
I think we should just remove this entire panel from the carousel, as everything on there is kinda contentious now.

Eilon:
These "images" are SVG files. In particular, `banner3.svg`, such as this instance: https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/wwwroot/images/banner3.svg

So I guess we should delete banner3.svg from everywhere, and rename banner4 to banner3.

Issue #330, opened 2018-02-20
Create and use new Third Party Notices URL for templates


Eilon:
Currently the 2.1 templates use a URL that was meant for earlier versions of the template, and links to the dotnet/templating repo. The ASP.NET Core templates need to have their own notices file with its own URL.

We need to:
* [ ] Create a Git **tag** for a version of https://github.com/aspnet/templating/blob/dev/src/THIRD-PARTY-NOTICES that has what we want for the 2.1.0 (previews, RC, RTM, should all be the same file) 
* [x] Create an AKA.MS link that points to it (I can do that)
* [x] Change the `thirdPartyNotices` metadata in the templates to point to the new AKA.MS link (here's an example of where it's used: https://github.com/aspnet/templating/blob/acc2147ac5b3092df194fd9ce043263b091ae4ae/src/Microsoft.DotNet.Web.ProjectTemplates/content/EmptyWeb-FSharp/.template.config/template.json#L14)

Eilon:
@jbagga - let's chat about this because we each need to do part of this work item.

Eilon:
Moving to RC1.

Eilon:
OK I created https://aka.ms/aspnetcore-template-3pn-210 , which right now redirects to https://github.com/aspnet/templating/blob/dev/src/THIRD-PARTY-NOTICES .

So, let's use that AKA.MS link for now, and then we'll update it at RTM time once we have a real tag for the release.

mkArtakMSFT:
We've looked through the PR during the Shiproom meeting yesterday. Go ahead with the commit, @jbagga.

jbagga:
See https://github.com/aspnet/templating/issues/480

PR #474, opened 2018-04-13
Update THIRD-PARTY-NOTICES link and remove Package Management panel from carousel

jbagga:
Addresses #330 and #469


src/Microsoft.DotNet.Web.ProjectTemplates/content/EmptyWeb-CSharp/.template.config/template.json
```diff
@@ -12,7 +12,6 @@
   "precedence": "3000",
   "identity": "Microsoft.Web.Empty.CSharp.2.1",
   "shortName": "web",
-  "thirdPartyNotices": "https://aka.ms/template-3pn",
```
jbagga: Some templates like Empty and WebAPI do not use any third party libraries and do not need the link. Safe to remove?

version.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.1.0</VersionPrefix>
-    <VersionSuffix>rc1</VersionSuffix>
+    <VersionPrefix>2.2.0</VersionPrefix>
```
mkArtakMSFT: Why is this `2.2` ?

jbagga: The PR was originally based off of dev before it was changed to 2.2.0. About to send an update to base it off of release/2.1 branch.

PR #476, opened 2018-04-15
update template to bs4

Eonasdan:
also upgrade libs to latest version. used cdnjs instead of the MS cdn since popper.js is missing (bs4 dependency) and because MS cdn doesn't show integrated strings.

I can makes changes to the other project templates if this looks good. The aspnet/Identity template would also be changed to target bs4 as the Manage screen in particular look wonky.

Please give me any feedback you have.


ryanbrandenburg:
We ended up wanting to change the general shape of our templates at the same time we moved to BS4, so we didn't end up using this. Thanks for the effort and feel free to submit anything you feel we missed as a new PR.

Issue #353, opened 2018-03-12
Update EF version in migrations in templates for RTM


ajcvickers:
See discussion here: https://github.com/aspnet/templating/pull/348

HaoK:
Note: Need to update updateMigrations.cmd to run dotnet ef

HaoK:
Done via https://github.com/aspnet/templating/pull/484

PR #481, opened 2018-04-19
Update migrations batch file + Review migration diffs

HaoK:
Part of https://github.com/aspnet/templating/issues/353

@ajcvickers so the automated updateMigrations.cmd is producing slightly different migrations than what we have checkedin, this PR is to review both the script (is there a way to specify the number in front of the schema?  Ideally I want it to use 000000000000 like we have hardcoded).

And the diffs... which appear to mostly be ordering in the designer


src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Data/SqlLite/00000000000000_CreateIdentitySchema.Designer.cs
```diff
@@ -1,20 +1,65 @@
...
 using Microsoft.EntityFrameworkCore.Infrastructure;
 using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Internal;
 using Microsoft.EntityFrameworkCore.Migrations;
 
 namespace Company.WebApplication1.Data.Migrations
 {
     [DbContext(typeof(ApplicationDbContext))]
-    [Migration("00000000000000_CreateIdentitySchema")]
+    [Migration("20180419233801_CreateIdentitySchema")]
```
HaoK: Is there anyway to hardcode this to 00000000000000_CreateIdentitySchema?

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Data/SqlLite/00000000000000_CreateIdentitySchema.Designer.cs
```diff
@@ -1,20 +1,65 @@
...
     partial class CreateIdentitySchema
     {
         protected override void BuildTargetModel(ModelBuilder modelBuilder)
         {
 #pragma warning disable 612, 618
             modelBuilder
-                .HasAnnotation("ProductVersion", "2.1.0-preview1");
+                .HasAnnotation("ProductVersion", "2.1.0-preview2-30571");
```
HaoK: We still need to close on what version string we want here, I guess we can just hand update this to 2.1.0-final/rtm or whatever we want as well

HaoK:
This makes the updates a little bit better, so as things currently stand, the two remaining warts are in the designer:
```
    [Migration("20180419233811_CreateIdentitySchema")]

                .HasAnnotation("ProductVersion", "2.1.0-preview2-30571")
```

Even if we have to hand edit those by hand x4, I guess its not too bad.

ajcvickers:
@bricelam A few things here:
* Looks like we're including an internal namespace in the snapshot now--is this expected?
* There's a whitespace change that doesn't look right
* Can you answer @HaoK's question on the timestamp?

@HaoK This is not for RC1, correct?

HaoK:
@ajcvickers no this is just prep work for 2.1 RTM or whatever the next milestone after rc1 is

src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Data/SqlLite/00000000000000_CreateIdentitySchema.Designer.cs
bricelam: Not from command-line. You can override the `IMigrationsIdGenerator` service...

bricelam:
It would be good to understand where the `.Internal` is coming from. AFAIK, we should be generating a lot less of these--not more.

The witespace change looks correct in the new code. Was the old one a manual-edit mistake?

HaoK:
@bricelam probably I think its been there a long time

bricelam:
Did you use 2.1.0-preview2 to generate? I suspect 2.1.0-rc1 will not generate `using ...Internal`

HaoK:
Its using whatever's in dev's templating repo right now, I can double check to see what version it is

HaoK:
Replaced by https://github.com/aspnet/templating/pull/484

ajcvickers:
@HaoK Didn't mean to approve this one--I have both tabs open and approved the wrong tab. 🐑

Issue #486, opened 2018-04-23
Fix point-in-time TFM mismatch in tests for 2.2.


natemcmaster:
https://github.com/aspnet/templating/pull/479 introduce a cyclical dependency on the CLI. We should update templates to test with the currently-building version of the shared framework, not the one bundled in the CLI.

cc @JunTaoLuo @jbagga @ryanbrandenburg

PR #487, opened 2018-04-23
Add workaround for mismatch between TFM and AspNetCore TargetFramewor…

JunTaoLuo:
…kVersions

This is a point in time problem and I will file an issue to remove this workaround.


JunTaoLuo:
Addresses https://github.com/aspnet/templating/issues/486

Issue #277, opened 2018-02-01
Switch to the more-reliable Selenium-based tests by using web-driver


jbagga:
To mitigate flaky failures we may want to consider adding retries to tests. The CI flakiness has gone down ever since we updated npm but this should still be considered for random local test failures.

cc @javiercn @SteveSandersonMS

javiercn:
Yes. I think that adding a global retry policy for the tests is a good way to improve the resiliency of these kind of tests. There are samples documenting how to do this with xUnit and it can help with selenium time outs. 

Another option is to drastically increment selenium timeouts (for lets say 2 minutes) at the cost of tests taking longer to fail. (It should not affect existing running tests).

mkArtakMSFT:
For the #237 issue make sure it is also resolved as part of these changes.

javiercn:
I showed them how our tests run in selenium the other day.

javiercn:
@ryanbrandenburg Is there anything to do here after the changes you already did?

ryanbrandenburg:
Yeah, we can re-open if we experience new/continued problems.

PR #493, opened 2018-04-24
[Design] Use RemoteWebDriver for selenium tests

jbagga:
Addresses #277 

**In this PR:**

1. Use RemoteWebDriver
2. Use Chrome by default
3. Option to use Firefox or Edge using environment variables
4. Turn off parallel test run for Edge - web driver does not support parallel selenium tests https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/5980485/
5. Use AssemblyFixture to set up and tear down the selenium-standalone server and keep test run parallel (sample used https://github.com/xunit/samples.xunit/tree/f4d955a52d3298150ad2bd211c65a3477e184413/AssemblyFixtureExample)

**Note:**
Travis needs installation and Appveyor needs installation and starting the selenium server in yaml files despite the fixture for the tests. 
I have tried running the tests on a clean machine and the fixture takes care of spinning up the server. We plan to move to dotnet-ci bot soon which would not require any configuration.

**Next steps:**

1. Logging for tests: Once the test run starts, there is no logging related to where in the process it is
2. Graceful handling of failures: In case of a failure or interruption in the test run, there needs to be clean up code that stops all the dotnet processes that had begun and close down web drivers and the selenium java process
3. Capture selenium server logs only for failures: May be useful to capture the selenium server logs to debug

### DIFF
--- a/test/Templates.Test/EmptyWebTemplateTest.cs
+++ b/test/Templates.Test/EmptyWebTemplateTest.cs
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Testing.xunit;
+using Templates.Test.Helpers;
 using Xunit;
 using Xunit.Abstractions;
 
+[assembly: AssemblyFixture(typeof(SeleniumServerFixture))]
 namespace Templates.Test
 {
     public class EmptyWebTemplateTest : TemplateTestBase
--- a/test/Templates.Test/Helpers/SeleniumServerFixture.cs
+++ b/test/Templates.Test/Helpers/SeleniumServerFixture.cs
@@ -1,0 +1,74 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.InteropServices;
+
+namespace Templates.Test.Helpers
+{
+    public class SeleniumServerFixture : IDisposable
+    {
+        private string _workingDirectory;
+        private Process _serverProcess;
+        private static readonly object _serverLock = new object();
+
+        public SeleniumServerFixture()
+        {
+            _workingDirectory = Directory.GetCurrentDirectory();
+            StartSeleniumStandaloneServer();
+        }
+
+        public void Dispose()
+        {
+            _serverProcess.Kill();
+            _serverProcess.Dispose();
+
+            // Find the java process running selenium-standalone
+            var childProcesses = Process.GetProcessesByName("java");
+            foreach (var childProcess in childProcesses)
+            {
+                childProcess.Kill();
+                childProcess.Dispose();
+            }
+        }
+
+        public void StartSeleniumStandaloneServer()
+        {
+            lock (_serverLock)
+            {
+                RunViaShell(_workingDirectory, "npm install -g selenium standalone");
+            }
+            lock (_serverLock)
+            {
+                RunViaShell(_workingDirectory, "selenium-standalone install");
+            }
+
+            // Starts a java process that runs the selenium server
+            _serverProcess = RunViaShell(_workingDirectory, "selenium-standalone start");
+        }
+
+        private static Process RunViaShell(string workingDirectory, string commandAndArgs)
+        {
+            var (shellExe, argsPrefix) = RuntimeInformation.IsOSPlatform(OSPlatform.Windows)
+                ? ("cmd", "/c")
+                : ("bash", "-c");
+
+            return Run(workingDirectory, shellExe, $"{argsPrefix} \"{commandAndArgs}\"");
+        }
+
+        private static Process Run(string workingDirectory, string command, string args = null, IDictionary<string, string> envVars = null)
+        {
+            var startInfo = new ProcessStartInfo(command, args)
+            {
+                UseShellExecute = false,
+                CreateNoWindow = true,
+                WorkingDirectory = workingDirectory
+            };
+
+            return Process.Start(startInfo);
+        }
+    }
+}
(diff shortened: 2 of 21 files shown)


jbagga:
Waiting for `Appveyor` and `Travis` to run successfully. I will request review once I make necessary changes

jbagga:
@SteveSandersonMS @javiercn Please take a look whenever you can!

test/Templates.Test/EmptyWebTemplateTest.cs
```diff
@@ -2,9 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Testing.xunit;
+using Templates.Test.Helpers;
 using Xunit;
 using Xunit.Abstractions;
 
+[assembly: AssemblyFixture(typeof(SeleniumServerFixture))]
```
SteveSandersonMS: This looks cool. Didn't know this existed. I think we'll copy this technique in the Blazor repo too!

jbagga: It may actually be possible to have it in our Common repo and use the same code!

test/Templates.Test/Helpers/SeleniumServerFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+            _workingDirectory = Directory.GetCurrentDirectory();
+            StartSeleniumStandaloneServer();
+        }
+
+        public void Dispose()
+        {
+            _serverProcess.Kill();
+            _serverProcess.Dispose();
+
+            // Find the java process running selenium-standalone
```
SteveSandersonMS: I don't quite understand this bit. It looks like on line 26 the `selenium-standalone` process was already killed. Is the issue that we also need to kill the process tree underneath it?

The logic here is probably fine for our use cases but it looks slightly dangerous in that it kills *all* the `java` child processes running on the machine, even if they are unrelated to the tests. Might it be enough to only kill the child processes of `_serverProcess`?

test/Templates.Test/Helpers/XunitExtensions/XunitTestAssemblyRunnerWithAssemblyFixture.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Templates.Test.Helpers.XunitExtensions
+{
+    public class XunitTestAssemblyRunnerWithAssemblyFixture : XunitTestAssemblyRunner
```
SteveSandersonMS: This all looks great. Definitely want to copy this technique.

test/Templates.Test/Helpers/XunitExtensions/XunitTestFrameworkExecutorWithAssemblyFixture.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class XunitTestFrameworkExecutorWithAssemblyFixture : XunitTestFrameworkExecutor
+    {
+        public XunitTestFrameworkExecutorWithAssemblyFixture(AssemblyName assemblyName, ISourceInformationProvider sourceInformationProvider, IMessageSink diagnosticMessageSink)
+            : base(assemblyName, sourceInformationProvider, diagnosticMessageSink)
+        {
+        }
+
+        protected override async void RunTestCases(IEnumerable<IXunitTestCase> testCases, IMessageSink executionMessageSink, ITestFrameworkExecutionOptions executionOptions)
+        {
+            using (var assemblyRunner = new XunitTestAssemblyRunnerWithAssemblyFixture(TestAssembly, testCases, DiagnosticMessageSink, executionMessageSink, executionOptions))
```
SteveSandersonMS: Super minor point: missing normal code-style braces around line 21

test/Templates.Test/Helpers/SeleniumServerFixture.cs
jbagga: Yes, the issue is that line 26 kills the process that started the selenium server but not the underlying java process.

Agreed. On the CI at least, there shouldn't be random java processes running but this is not ideal for sure. The code to get child processes is a little round about but there is a way. Let me try it out and update the PR!

test/Templates.Test/Helpers/XunitExtensions/XunitTestFrameworkExecutorWithAssemblyFixture.cs
jbagga: Oops, missed it. Btw, most of this code is from this sample: https://github.com/xunit/samples.xunit/tree/f4d955a52d3298150ad2bd211c65a3477e184413/AssemblyFixtureExample

test/Templates.Test/Helpers/SeleniumServerFixture.cs
```diff
@@ -0,0 +1,103 @@
...
+            {
+                childProcess.Kill();
+                childProcess.Dispose();
+            }
+
+            _serverProcess.Kill();
+            _serverProcess.Dispose();
+        }
+
+        private static List<Process> GetChildProcesses(int parentProcessId)
```
jbagga: @SteveSandersonMS Thoughts on this? This gets all the children processes of the `_serverProcess`

jbagga: This is not cross platform. Doesn't work on Linux or Mac

jbagga:
Once https://github.com/aspnet/Coherence-Signed/issues/856 is done, run https://github.com/aspnet/templating/pull/519 on the CI. Then this can be merged.
I was unable to find a cross-platform and reliable way to programmatically shutdown the selenium-standalone server without killing all java processes. @SteveSandersonMS may have other ideas on how to handle that.

Thanks!

mkArtakMSFT:
@javiercn, can you please handle merging this PR in?

ryanbrandenburg:
I used most of this PR in my giant templating test fix PR, thanks @jbagga!

PR #495, opened 2018-04-26
Import styles in correct order

fredrik-lundin:
In the `angular-cli.json` file, the styles.css file should be imported last in the styles section, so the developer can override styles from third party libraries (bootstrap, in this case).

This change will make it so that custom styles will be added after the bootstrap styles.


fredrik-lundin:
Anyone care to take a look? :) It's a really small, but important change

@SteveSandersonMS @jbagga @HaoK

mkArtakMSFT:
Sorry for taking so long to take this in, @fredrik-lundin.
@ryanbrandenburg, can you please look into this?

mkArtakMSFT:
Also, @fredrik-lundin, can you please resolve the conflict (just update) and we'll take this in quickly.

mkArtakMSFT:
@ryanbrandenburg this is going to be merged into master. Should this be ported to release/2.2 also? If that's the case, can you please take care of it? Thanks!

ryanbrandenburg:
Considering that we were re-targeting the branch it seemed easiest to create a new PR. #668 is merged into release/2.2 now, the bot will create a PR to merge forward into master. Thanks again @fredrik-lundin!

PR #508, opened 2018-05-03
Upgrade to netcoreapp22

ryanbrandenburg:


ryanbrandenburg:
I will be on vacation, but I believe that this PR will pass if, the next time Universe publishes, you run `./run.ps1 -update upgrade deps` on it. It just needs a fixed version of Common.

pranavkm:
Tests passed, but the build failed due to https://github.com/aspnet/templating/issues/514 which was observed prior to the TFM change. Getting this PR in to get us moving.

PR #524, opened 2018-05-17
 Added tests to verify fallback source content is same as the CDN content

kichalla:


src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/wwwroot/lib/jquery/dist/jquery.min.js
```diff
@@ -1,9 +1,2 @@
-<<<<<<< HEAD
```
natemcmaster: 🤦‍♂️ let's open a bug to fix this in the 2.1.x patches as well.

kichalla: Sure, will do

PR #533, opened 2018-05-22
Remove reference to Microsoft.AspNetCore.Mvc.Razor.ViewCompilation fr…

### LINKED ISSUE #530: Remove reference to Microsoft.AspNetCore.Mvc.Razor.ViewCompilation from desktop templates
For instance - https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/StarterWeb-CSharp.csproj.in#L44

ViewCompilation isn't recommended any longer, and we shouldn't reference it from our templates.

pranavkm:
…om desktop templates

Fixes #530


pranavkm:
Leaving it in the F# templates because Razor Sdk does not apply there.

PR #534, opened 2018-05-22
CompatibilityVersion -> 2.2

ryanbrandenburg:
We shouldn't merge this until aspnet/Mvc#7812 goes in. @pranavkm we definitely want Version_2_2 and not Latest right?


pranavkm:
Should be good to merge now

PR #539, opened 2018-05-27
Fixing issue in the react and redux templates where the test is failing

lancscoder:
Currently the javascript test in the React and Redux templates is failing with the following error:

```
 console.error node_modules\fbjs\lib\warning.js:33
   Warning: Failed context type: The context `router` is marked as required in `Link`, but its value is `undefined`.
       in Link (at NavMenu.js:11)
       in NavbarBrand (at NavMenu.js:10)
       in div (created by NavbarHeader)
       in NavbarHeader (at NavMenu.js:9)
       in div (created by Grid)
       in Grid (created by Navbar)
       in nav (created by Navbar)
       in Navbar (created by Uncontrolled(Navbar))
       in Uncontrolled(Navbar) (at NavMenu.js:8)
       in Unknown (at Layout.js:9)
       in div (created by Col)
       in Col (at Layout.js:8)
       in div (created by Row)
       in Row (at Layout.js:7)
       in div (created by Grid)
       in Grid (at Layout.js:6)
       in Unknown (at App.js:9)
       in Unknown (at App.test.js:7)
```

This is because in the test we're trying to render `<App />` with no context of the router. Therefore when trying to render a link the above error is thrown.

### DIFF
--- a/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.test.js
+++ b/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.test.js
@@ -1,8 +1,12 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
+import { MemoryRouter } from 'react-router-dom';
 import App from './App';
 
 it('renders without crashing', () => {
   const div = document.createElement('div');
-  ReactDOM.render(<App />, div);
+  ReactDOM.render(
+    <MemoryRouter>
+      <App />
+    </MemoryRouter>, div);
 });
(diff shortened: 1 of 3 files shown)


ryanbrandenburg:
Actually, on second read-through I noticed that this is our first use of `redux-mock-store`, I believe we'll have to get approval to use it before we can merge this.

ryanbrandenburg:
So, when I went a little deeper here I realized that we can avoid the `redux-mock store` package entirely with a couple of lines, so I created #554. Would you agree that's equivalent? If so we can either merge #554 or you can make a similar change to this PR and we'll merge this. Whichever you prefer. Either way, thanks!

lancscoder:
@ryanbrandenburg yes - I would agree the change you've made in  #554 is equivalent. 

I've updated the PR. Thank

Issue #525, opened 2018-05-17
[Patch for 2.1.x] Fix the jquery.min.js file in RazorPages web application


kichalla:
Original issue: https://github.com/aspnet/templating/issues/523

kichalla:
The following commit, from 'dev' branch, needs to be cherry picked onto the patch branch:
https://github.com/aspnet/templating/commit/90fe58be756b3bc96d527bbdc8a03ce30b56080f

mkArtakMSFT:
@kichalla, I assume this is ready to be closed now, is it?

PR #543, opened 2018-05-30
[Fixes #523] Fix the jquery.min.js file in RazorPages web application

### LINKED ISSUE #523: Fix the jquery.min.js file in RazorPages web application
https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/wwwroot/lib/jquery/dist/jquery.min.js

This issue is also present in `release/2.1`

kichalla:
https://github.com/aspnet/templating/issues/525


natemcmaster:
Should we also port your SRI hash checks tests?

kichalla:
> Should we also port your SRI hash checks tests?

Sure, will do.

PR #564, opened 2018-06-11
ResponseCache attribute on RazorPages Error handler method should be …

### LINKED ISSUE #553: ResponseCache attribute on RazorPages Error handler method should be moved to model
See https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Pages/Error.cshtml.cs#L17-L21

Filters cannot be applied to handler methods, only to the model type.

pranavkm:
…moved to model

Fixes #553


pranavkm:
FYI @kichalla

PR #575, opened 2018-06-18
Fix #574: Set NoDefaultLaunchSettingsFile

### LINKED ISSUE #574: --exclude-launch-settings should set NoDefaultLaunchSettingsFile
When you create a new project using the `--exclude-launch-settings` option, the generated template will not contain a `Properties/launchSettings.json` files. However, as soon as you launch Visual Studio, it will automatically create a `launchSettings.json` file since that’s the default behavior with Visual Studio.

So when one enables `ExcludeLaunchSettings`, the following setting should be added to the generated `.csproj`:

    <NoDefaultLaunchSettingsFile>True</NoDefaultLaunchSettingsFile>

This disables the default launch settings behavior and will prevent Visual Studio from generating one.

---

If I figure out how the template’s project files are generated, I’ll attempt to provide a PR for this.

poke:
This is an attempt to fix #574.

I have no idea if this is all that’s necessary to make this work. ~~~Unfortunately, I wasn’t able to build and the test the templates correctly. Running the build.cmd completely messed with my environment…~~~ Unfortunately, I have no idea how to use the templates after installing them without breaking my existing environment by uninstalling the stable template packages… :/


poke:
Rebased against current dev. If this gets accepted, I’ll also provide a PR against 2.2 (or even 2.1 if desired).

poke:
@mkArtakMSFT Sure thing and done :)

PR #582, opened 2018-06-20
[Design]Update CI configs to use node 10

ryanbrandenburg:


ryanbrandenburg:
@SteveSandersonMS is that all there is to it? I feel like I must be missing something, otherwise the only work here is on the build/only end.

SteveSandersonMS:
@ryanbrandenburg It's totally believable to me that this is all you need. There shouldn't be any code changes needed, since Node 10 should easily be back-compatible with (say) Node 8 for the cases we need.

ryanbrandenburg:
Upgrading to node 10 seems to [break](https://travis-ci.org/aspnet/templating/jobs/397511005#L1600) [angular](https://github.com/angular/angular-cli/issues/9283) for mac. The suggested fixes seem to be 1) Don't check in package-lock.json files. 2) Use a version of npm that won't complain.

I'm not sure it's worth it to solve this problem. So far as I know there's no product reason why we should use node 10 over node 8, and it's a build-time only change anyway. @SteveSandersonMS thoughts?

SteveSandersonMS:
I agree that there's no pressing reason to use Node 10 in CI, so if it's not stable yet I suggest we stay on v8. Thanks for looking into it though!

New versions of Node often aren't stable, at least in the sense that NPM doesn't usually ship a compatible new version on the same day. We should wait it out until this is resolved.

> Don't check in package-lock.json files

We definitely *should* include `package-lock.json` in source control, otherwise we're not getting any benefit from `package-lock.json` at all. The point of it is to ensure deterministic restores rather than picking up whatever random collection of transitive dependencies happens to be latest on each restore.

PR #602, opened 2018-07-02
Resolve build warnings about missing template properties

natemcmaster:
It adddresses the build warnings only. Tests were still failing on my machine.


ryanbrandenburg (approved):
:shipit:

Issue #565, opened 2018-06-11
Port ResponseCache attribute on RazorPages Error handler method fix to patch


pranavkm:
Original issue: https://github.com/aspnet/templating/issues/553

Commit to port: https://github.com/aspnet/templating/commit/e403187be35a25560692aaa7ded495e26bb3ad90

pranavkm:
cc @mkArtakMSFT

muratg:
Approved for 2.1.3.

ryanbrandenburg:
Fixed in #609.

PR #609, opened 2018-07-11
ResponseCache attribute on RazorPages Error handler method should be …

### LINKED ISSUE #553: ResponseCache attribute on RazorPages Error handler method should be moved to model

ryanbrandenburg:
…moved to model

Fixes #553 

Porting this back to 2.1.3 as per #565.


ryanbrandenburg:
Merging as per @muratg's advice. To the best of my knowledge the failures are pre-existing conditions, and I tested some of the sites locally.

PR #618, opened 2018-07-17
Rename Web Application to Starter Web Application

### LINKED ISSUE #584: Rename Web Application templates to Starter Web Application

ryanbrandenburg:
Fixes #584. @danroth27 to confirm that we just wanted to change the display-name, not the shortName, which would be what you typed to create a template. If the shortName need to change too I just need to confirm what it would be called. We could alias it like so:
```json
"shortName": [
  "mvc",
  "starter-mvc"
]
```


danroth27:
Yup, we just wanted to change the display name.

PR #659, opened 2018-08-06
Removed the `layout=null` assignment from the page

mkArtakMSFT:


mkArtakMSFT:
Verified that not having assignment allows the layout to be overridden by the main app.

PR #660, opened 2018-08-06
Experience review feedback

ryanbrandenburg:


danroth27:
What does the template look like on a wide monitor? Does the width of the margins looks reasonable? 

Can we also make the content in the navbar align with the content and margins in the body?

PR #671, opened 2018-08-10
Add .gitattributes file

### LINKED ISSUE #408: Add missing .gitattributes to this repo

ryanbrandenburg:
Fixes #408.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,52 @@
+*.doc  diff=astextplain
+*.DOC	diff=astextplain
+*.docx	diff=astextplain
+*.DOCX	diff=astextplain
+*.dot	diff=astextplain
+*.DOT	diff=astextplain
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,52 @@
+*.doc  diff=astextplain
+*.DOC	diff=astextplain
+*.docx	diff=astextplain
```
ajaybhargavb: Why whitespace so inconsistent?

ryanbrandenburg: This is just a direct copy of what we have in [aspnet/Mvc](https://github.com/aspnet/Mvc/blob/release/2.2/.gitattributes). Which [other repos](https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/.gitattributes) also directly ripped off.

PR #672, opened 2018-08-10
Don't publish too much of ClientApp

ryanbrandenburg:


ryanbrandenburg:
CC @mikeharder who noticed this.

mikeharder:
@ryanbrandenburg: You might also need to revert:

https://github.com/aspnet/Templating/commit/d95f9716932cae55933c0a4700fec3513c55c92f#diff-7d3194c7a47f8f21106bbc7fe3c1ee29L56

ryanbrandenburg:
I tested it manually on my machine and we also have tests of the [publish scenario](https://github.com/aspnet/Templating/blob/release/2.2/test/Templates.Test/Helpers/AspNetProcess.cs#L30) so I'd say we're save if you don't have any concerns.

mikeharder:
@ryanbrandenburg : No concerns from my side.

Issue #471, opened 2018-04-13
Update project templates to use LibMan


DamianEdwards:
LibMan now ships in VS 15.7. It will soon be followed by a CLI tool that will enable it for cmd line scenarios. When that happens, we need to update the templates to use LibMan for managing client dependencies. This will involve including a `libman.json` file in the root of the projects, configured to bring in the required client libs (jQuery, jQuery Validate, jQuery Validation Unobtrusive, and Bootstrap).

Note this is blocked on getting confirmation from @ChrisSfanos RE approval/requirements for using cdnjs as the configured source in VS and templates.

guardrex:
For our upcoming docs issue(s), is it likely that LibMan will become part of the templates for 2.2?

DamianEdwards:
Very, possibly earlier (in a 2.1.x patch)

danroth27:
LibMan was pulled from VS2017 15.7, because it wasn't quite ready yet, but it's still being actively worked on so we should keep this bug to track the template integration.

PR #679, opened 2018-08-17
Added support for LibMan and removed old banner images

JustCallMeAD:
Modified the starter template so that it supports LibMan by default. Also deleted the old banner images as they are not being used anymore.


JustCallMeAD:
NOTE: Will need to make similar changes to the identity UI templates. I can do a pull request for these changes too if you would like me too?

src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/libman.json
```diff
@@ -0,0 +1,57 @@
+{
+  "version": "1.0",
+  "defaultProvider": "cdnjs",
+  "libraries": [
+    {
+      "provider": "cdnjs",
```
scottaddie: You can remove the `provider` property from each of these object literals. The `defaultProvider` property will be used.

scottaddie:
@justcla You may be interested in this PR.

Eilon:
@mkArtakMSFT / @ryanbrandenburg - please check with @DamianEdwards on when/how we should get this in.

ryanbrandenburg:
[Seems like](https://github.com/aspnet/Templating/issues/471#event-1683852954) we're targeting LibMan for 3.0.0-preview1.

ryanbrandenburg:
Please rebase onto the latest from master. I ended up removing the banner images independently and making other changes in the mean time and there's a conflict in StarterWeb's _Layout.cshtml, so it will be much easier to see what's actually changing once this PR has the latest.

ryanbrandenburg:
@JustCallMeAD, a reminder to rebase this onto master so we can more easily see what the important changes are.

ryanbrandenburg (changes requested):
I poked around at this locally and I think there are some extra things we'll want to address when doing this:

1. We should ensure that LibMan is installed, then run `libman restore` before the build step. A target similar to https://github.com/aspnet/Templating/blob/master/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/Angular-CSharp.csproj.in#L40 would do the trick.
2. Let's not check the files that LibMan pulls down for us into Git anymore (they'll just get overridden anyway). That will break our baseline tests, so we'll have to update those and also confirm that all scenarios still work, integrity attributes still apply etc.

Since the LibMan work is scheduled for 3.0 (and we're still doing 2.2 work) either you can do those cleanup items or it'll be a bit of a wait before I can carve out some time. Either way this is a pretty good start and probably what we'll be looking to for #471.

Eilon:
Hi, we are closing this PR because the source code has all been moved to https://github.com/aspnet/AspNetCore/tree/master/src/Templating so that all the ASP.NET Core source code is in one place. If you are interested in continuing this contribution, we recommend you log a new issue in https://github.com/aspnet/AspNetCore/issues with details and if the proposal is accepted, send a new PR so the new repo. Sorry for any inconvenience!

PR #688, opened 2018-08-20
Add comments regarding HSTS max age

### LINKED ISSUE #655: Include HSTS configuration when using app.UseHsts();
https://github.com/aspnet/BasicMiddleware/pull/354

The new ASP.NET Core 2.1 templates are including `app.UseHsts()`, which is adding the Strict-Transport-Security header. Which is awesome. 
But the default for the max-age parameter in Microsoft.AspNetCore.HttpsPolicy is 30 days to address initial configuration issues when deploying a new app. The configuration of this is not that straightforward, so we should add a configuration code (or a comment at least). Now we are encouraging developers to leave it on 30 days which is not that great...

```
// Change max-age to a longer period (1 year at least) after you are confident with your configuration
services.Configure<HstsOptions>(options =>
{
    options.IncludeSubDomains = true;
    options.MaxAge = TimeSpan.FromDays(30);
});
```

ryanbrandenburg:
Fixes #655.

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
@@ -147,6 +147,7 @@ namespace Company.WebApplication1
             {
                 app.UseExceptionHandler("/Error");
 #if (RequiresHttps)
+                // The default HSTS value is 30 days. You may want to change this for production scenarios, see https://aka.ms/aspnetcore-hsts.
                 app.UseHsts();
             }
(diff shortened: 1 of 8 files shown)


blowdart:
LGTM

guardrex:
FYI: We **do** document the 30-day default in the docs.

Referring to example code that sets the default to 60, we then say ...

> Explicitly sets the max-age parameter of the `Strict-Transport-Security` header to to 60 days. If not set, defaults to 30 days. See the `max-age` directive for more information.

https://docs.microsoft.com/aspnet/core/security/enforcing-ssl#http-strict-transport-security-protocol-hsts

**_We're good!_** :+1:

PR #698, opened 2018-08-21
Update precedence for 2.2 templates

javiercn:
Updates the precedence for 2.2 templates.
When 2.1 and 2.2 templates are installed on the same machine and the templating engine can't differentiate based on the provided parameters the 2.2 templates will be selected.


javiercn:
3ed89516bae1c9b36fc1c2c800ad0111cd95563b

Issue #692, opened 2018-08-21
Incorrect page links in the Bootstrap 4 Razor Pages based template


danroth27:
The links to Page1, Page2, and the Privacy page should be based on pages, not actions and controllers:

*_Layout.cshtml*
```html
<li class="nav-item">
    <a class="nav-link text-dark" asp-area="" asp-controller="Home" asp-action="Index">Page1</a>
</li>
<li class="nav-item">
    <a class="nav-link text-dark" asp-area="" asp-controller="Home" asp-action="Index">Page2</a>
</li>

...

<footer class="border-top footer pl-3 text-muted">
    &copy; Year - WebApplication20 - <a asp-area="" asp-page="Home" asp-action="Privacy">Privacy</a>
</footer>
```

ryanbrandenburg:
This and some other issues are fixed in https://github.com/aspnet/Templating/pull/699.

PR #699, opened 2018-08-21
Fix template layouts

ryanbrandenburg:
Fixes aspnet/Templating#692. The fact that that issue happened made me decide to walk through each template again, and I found several other issues with basic appearance and layouts.

To view and test these yourself, simple run `.\scripts\Run-<SomeTemplate>.ps1`. If that script doesn't work out of the box either leave a comment here or file an issue.


ryanbrandenburg:
Agreed. @mkArtakMSFT put something on the calendar for us to do that monday of next week.

ryanbrandenburg:
I went over this with @danroth27 (today) and @javiercn (Monday) and both approved so I'm going to merge to avoid coming in at the last minute for preview2. If anyone has any objections that weren't voiced here let me know.

PR #701, opened 2018-08-22
Fixes layout issues and login partial issues

javiercn:
Validated 
* MVC individual
* Razor pages individual
* MVC B2C auth
* Razor B2C auth

### DIFF
--- a/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Views/Shared/_LoginPartial.Identity.cshtml
+++ b/src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Views/Shared/_LoginPartial.Identity.cshtml
@@ -1,22 +1,26 @@
+@using Microsoft.AspNetCore.Identity
+@inject SignInManager<IdentityUser> SignInManager
+@inject UserManager<IdentityUser> UserManager
+
 <ul class="navbar-nav">
-    @if (User.Identity.IsAuthenticated)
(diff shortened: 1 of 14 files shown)


src/Microsoft.DotNet.Web.ProjectTemplates/content/StarterWeb-CSharp/Views/Shared/_LoginPartial.Identity.cshtml
```diff
@@ -1,22 +1,26 @@
...
-        <li class="nav-item">
-            <form id="logoutForm" class="form-inline" asp-area="Identity" asp-page="/Account/Logout" asp-route-returnUrl="@Url.Page("/Index", new { area = "" })">
-                <button id="logout" type="submit" class="nav-link btn btn-link text-dark">Logout</button>
-            </form>
-        </li>
-    }
-    else
-    {
-        <li class="nav-item">
-            <a class="nav-link text-dark" id="register" asp-area="Identity" asp-page="/Account/Register">Register</a>
```
ryanbrandenburg: Yeah, I wondered about these ID attributes. Is it generally considered good practice to remove id's and names when they aren't used by anything?

javiercn: The code that I gave you had id attributes because we used them for functional tests, but we shouldn't need to do that

PR #335, opened 2018-02-28
Remove bundleconfig.json files from project templates

jbagga:
Addresses #326

PR #729, opened 2018-09-07
Normalize FSharp templates

### LINKED ISSUE #724: Bring F# templates in line with other templates
We need to make the F# templates match their C# counterparts. It should mostly just be a matter of copying cshtml files.

ryanbrandenburg:
Fixes #724.

`site.min.*` files are also no longer checked in, but are generated at build time to be included in the packages.


build/repo.targets
```diff
@@ -0,0 +1,18 @@
...
+        <Exec Command="node --version" ContinueOnError="true">
+            <Output TaskParameter="ExitCode" PropertyName="ErrorCode" />
+        </Exec>
+        <Error Condition="'$(ErrorCode)' != '0'" Text="Node.js is required to build and run this project. To continue, please install Node.js from https://nodejs.org/, and then restart your command prompt or IDE." />
+        <ItemGroup>
+            <SiteJsFiles Include="$(RepositoryRoot)src\Microsoft.DotNet.Web.ProjectTemplates\content\*\wwwroot\js\site.js" />
+            <SiteCssFiles Include="$(RepositoryRoot)src\Microsoft.DotNet.Web.ProjectTemplates\content\*\wwwroot\css\site.css" />
+        </ItemGroup>
+        <Exec Command="npx uglifycss %(SiteCssFiles.Identity) --output %(RelativeDir)site.min.css" />
+        <Exec Command="npx uglify-js %(SiteJsFiles.Identity) --output %(RelativeDir)site.min.js" />
```
SteveSandersonMS: I'd imagine this is not super fast, as it's creating a new Node process for every `.css`/`.js` file in all the templates. It also relies on `npx` being installed globally, which is not standard within a Node environment.

Maybe it's OK, but if either aspect of this becomes problematic, we could add a small Node script that does it (so, one Node process that iterates through all the files and minifies them) and have it invoked via `node ourscript.js`.

SteveSandersonMS:
Update: I see that @Eilon also raised the concern about nothing updating the `.min.*` files previously ([here](https://github.com/aspnet/Templating/pull/335#issuecomment-369425021)).

It seems like the outcome was to "keep the file there to keep the pattern going" (from [here](https://github.com/aspnet/Templating/pull/335#issuecomment-369439204)). To me it doesn't seem like a good enough justification for creating a trap for people to fall into, but maybe there was more to the conversation than was summarised there.

ryanbrandenburg:
I agree with you for our current state, but I believe the plan is to start managing that using LibMan in the near future. @danroth27 or @DamianEdwards is that accurate (or even possible)?

ryanbrandenburg:
I talked to @Eilon and @DamianEdwards in person and we agreed to kill `.min.css` files. I'll change things here to reflect that.

ryanbrandenburg:
:up::date:ed to remove all `site.min.*`.

ryanbrandenburg:
For posterity, apparently LibMan doesn't/wont do minification, not sure where I got the idea that it could.

SteveSandersonMS:
@ryanbrandenburg Looks superb!

PR #751, opened 2018-09-24
Don't use depricated methods

ryanbrandenburg:


SteveSandersonMS:
This diff is unexpected for a few reasons:

* It populates the component's `state` even though:
    * Under the Redux model, this particular bit of state is stored externally in the Redux store, not privately by the component
    * The component doesn't appear to read back its own `state` data anyway
* It defines a `handleChange` property that never gets used
* The `render` function causes side-effects (specifically, triggering `requestWeatherForecasts` on the Redux store)

An alternate approach would be retaining most of the existing structure, but just swapping `componentWillMount`/`componentWillReceiveProps` with the non-deprecated `componentDidMount`/`componentDidUpdate`. Since the parameter to `componentDidUpdate` is `prevProps` rather than `nextProps`, the implementation changes slightly. Example:

```javascript
import React, { Component } from 'react';
import { bindActionCreators } from 'redux';
import { connect } from 'react-redux';
import { Link } from 'react-router-dom';
import { actionCreators } from '../store/WeatherForecasts';

class FetchData extends Component {
  componentDidMount() {
    // This method is called when the component is first added to the document
    this.ensureDataFetched();
  }

  componentDidUpdate() {
    // This method is called when the route parameters change
    this.ensureDataFetched();
  }

  ensureDataFetched() {
    const startDateIndex = parseInt(this.props.match.params.startDateIndex, 10) || 0;
    this.props.requestWeatherForecasts(startDateIndex);
  }

  render() {
    return (
      <div>
        <h1>Weather forecast</h1>
        <p>This component demonstrates fetching data from the server and working with URL parameters.</p>
        {renderForecastsTable(this.props)}
        {renderPagination(this.props)}
      </div>
    );
  }
}

function renderForecastsTable(props) {
  return (
    <table className='table table-striped'>
      <thead>
        <tr>
          <th>Date</th>
          <th>Temp. (C)</th>
          <th>Temp. (F)</th>
          <th>Summary</th>
        </tr>
      </thead>
      <tbody>
        {props.forecasts.map(forecast =>
          <tr key={forecast.dateFormatted}>
            <td>{forecast.dateFormatted}</td>
            <td>{forecast.temperatureC}</td>
            <td>{forecast.temperatureF}</td>
            <td>{forecast.summary}</td>
          </tr>
        )}
      </tbody>
    </table>
  );
}

function renderPagination(props) {
  const prevStartDateIndex = (props.startDateIndex || 0) - 5;
  const nextStartDateIndex = (props.startDateIndex || 0) + 5;

  return <p className='clearfix text-center'>
    <Link className='btn btn-default pull-left' to={`/fetch-data/${prevStartDateIndex}`}>Previous</Link>
    <Link className='btn btn-default pull-right' to={`/fetch-data/${nextStartDateIndex}`}>Next</Link>
    {props.isLoading ? <span>Loading...</span> : []}
  </p>;
}

export default connect(
  state => state.weatherForecasts,
  dispatch => bindActionCreators(actionCreators, dispatch)
)(FetchData);
```

From what I can tell, this uses future-supported APIs and behaves correctly.

**Separate issue**

While investigating this, I also noticed a more major issue, which is that since the Bootstrap 4 redesign, the nav links at the top of the page don't trigger client-side navigation any more - they now trigger a complete page reload, defeating one of the main purposes of having a SPA.

This is because it now uses `<NavbarBrand>` and `<NavLink>` instead of `<Link>`. To fix this, could you update the `NavMenu.js` to have the following `render` function?

```javascript
  // Also add at the top: import { Link } from 'react-router-dom';

  render() {
    return (
      <header>
        <Navbar className="navbar-expand-sm navbar-toggleable-sm border-bottom box-shadow mb-3" light >
          <Container>
            <NavbarBrand tag={Link} to='/'>Company.WebApplication1</NavbarBrand>
            <NavbarToggler onClick={this.toggle} className="mr-2" />
            <Collapse className="d-sm-inline-flex flex-sm-row-reverse" isOpen={this.state.isOpen} navbar>
              <ul className="navbar-nav flex-grow">
                <NavItem>
                  <NavLink tag={Link} className="text-dark" to="/">Home</NavLink>
                </NavItem>
                <NavItem>
                  <NavLink tag={Link} className="text-dark" to="/counter">Counter</NavLink>
                </NavItem>
                <NavItem>
                  <NavLink tag={Link} className="text-dark" to="/fetch-data">Fetch data</NavLink>
                </NavItem>
              </ul>
            </Collapse>
          </Container>
        </Navbar>
      </header>
    );
  }
```

If it's unclear what I'm saying is wrong, please let me know!

ryanbrandenburg:
Good catch on both accounts! The page feels very smooth once we don't do a full refresh, I'll put that into this PR.

ryanbrandenburg:
:up::date:

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/ReactRedux-CSharp/ClientApp/src/components/FetchData.js
```diff
@@ -5,19 +5,25 @@ import { Link } from 'react-router-dom';
...
-    // This method runs when incoming props (e.g., route params) change
-    const startDateIndex = parseInt(nextProps.match.params.startDateIndex, 10) || 0;
+  ensureDataFetched() {
+    const startDateIndex = parseInt(this.props.match.params.startDateIndex, 10) || 0;
     this.props.requestWeatherForecasts(startDateIndex);
   }
 
   render() {
+    const startDateIndex = parseInt(this.props.match.params.startDateIndex, 10) || 0;
+    this.props.requestWeatherForecasts(startDateIndex);
```
SteveSandersonMS: I don't think lines 24 and 25 need to be added. The call to `requestWeatherForecasts` is handed in `ensureDataFetched` so doesn't need to be repeated in `render`. In general it's desirable for `render` functions to be free of side-effects.

When you remove these two lines (and the following whitespace one I guess), it will still work, won't it?

ryanbrandenburg:
Yep it still worked after removing it from rendering (I just missed it when cleaning up). Merging.

PR #756, opened 2018-10-02
Updated react-scripts dependency for the React-CSharp template to version 2.0.3

mauricedb:
React-Scripts just received a major overhaul and generates a lightly different React application. This PR updates the dependencies and changes in service worker registration.

### DIFF
--- a/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/package-lock.json
+++ b/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/package-lock.json
@@ -5,154 +5,1005 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.51",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.51.tgz",
-      "integrity": "sha1-vXHZsZKvl435FYKdOdQJRFZDmgw=",
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "requires": {
-        "@babel/highlight": "7.0.0-beta.51"
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/core": {
+      "version": "7.1.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.1.0.tgz",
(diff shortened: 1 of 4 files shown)


Eilon:
@SteveSandersonMS thoughts on this?

SteveSandersonMS:
@Eilon I agree this would be a great thing to do. Our intent is definitely to keep at least reasonably up-to-date with the underlying SPA technologies.

As for actually merging this PR, I think as a matter of due diligence (and hopefully no offence to @mauricedb - thanks for the contribution!), one of us on the team should also go through the process of regenerating the React app from the latest CRA, working out the diffs versus our template, and making sure it aligns with @ryanbrandenburg's work on the latest Bootstrap 4-based design. This is because there's no other way to verify that all the changes are included and we are responsible for any differences versus the stock CRA output. I can't know just by reading the diff here whether there might be any other diffs versus the latest CRA output. Fortunately @mauricedb's PR suggests the work here will be very minor, as the only nontrivial change is in `serviceWorker.js`, which is a file we don't have any customizations to, so we can just use their latest version of that file as generated.

Can this work be added to some suitable backlog? If @ryanbrandenburg is continuing template ownership, I'm sure this will be a very straightforward job for him, though I guess it's more for @mkArtakMSFT to decide who's assigned.

Ideally when we do investigate, we'll discover that no further changes are needed versus @mauricedb's PR and we can just merge this as-is.

mauricedb:
@SteveSandersonMS No offence taken, if you want to take a different approach just ask for me to make changes or reject my PR and open a new one :-)

SteveSandersonMS:
@mkArtakMSFT Rather than being a PR we can review directly, I'd see this more as a regular backlog work item as in [my comment before](https://github.com/aspnet/Templating/pull/756#issuecomment-426571175). Please ping me if this seems strange or it's not clear what I'm suggesting we do :)

mkArtakMSFT:
Thanks @SteveSandersonMS. Agree, that this should go through a validation. @ryanbrandenburg, let's try to do this during 2.2.

ryanbrandenburg:
@SteveSandersonMS as best I can tell the WebSocket warning is coming out of JavaScriptServices. Can you shed any light on this?

SteveSandersonMS:
@ryanbrandenburg I believe we reverse-proxy the websocket connection to the underlying SPA dev server: https://github.com/aspnet/JavaScriptServices/blob/master/src/Microsoft.AspNetCore.SpaServices.Extensions/Proxying/SpaProxy.cs. Perhaps something in our proxying logic doesn't like something about the request.

ryanbrandenburg:
This is a timing issue of come kind, because it goes away if I add a breakpoint in websocket.js.

Eilon:
Hi, we are closing this PR because the source code has all been moved to https://github.com/aspnet/AspNetCore/tree/master/src/Templating so that all the ASP.NET Core source code is in one place. If you are interested in continuing this contribution, we recommend you log a new issue in https://github.com/aspnet/AspNetCore/issues with details and if the proposal is accepted, send a new PR so the new repo. Sorry for any inconvenience!

PR #779, opened 2018-10-15
Integrity

ryanbrandenburg:


javiercn:
As pointed out in the bug, can you provide a description on why we are doing this? A 5 min google search didn't do the trick

PR #794, opened 2018-10-22
Handle peer dependency warnings

### LINKED ISSUE #784: npm ERR! peer dep missing
To my knowledge the messages [as seen here](https://github.com/aspnet/Release/issues/322#issuecomment-430387094) have been this way for at least a couple releases, but it would be nice to either add these dependencies or silence the warnings depending on which is cleaner/better.

ryanbrandenburg:
Fixes #784.

Basically the problem here for all the angular stuff was that `@angular/platform-server` was 6.0.0 and the packages that would normally be it's peer dependencies were `^6.0.0`. Those packages resolved to 6.0.5, but platform-server wanted exactly 6.0.0, so it complained.

We also added jquery and popper.js to silence warnings from bootstrap.

New versions and packages will require OSS review from @Eilon.

CC @dougbu


ryanbrandenburg:
`Custom-Hive.ps1` changed because @javiercn pointed out an improvement I could make to the script to make it more robust and simpler. I found it useful when checking the Angular changes locally, so I kept it in.

SteveSandersonMS:
As just discussed in person (but posting here to explain to others), we're hoping this could be solved not by *downgrading* all the Angular packages from 6.0.5 to 6.0.0, but instead by *upgrading* platform-server (and all the others for consistency) to the latest in the 6.x line. Currently that's 6.1.10.

If it turns out that 6.1.10 causes us any problems, we could use all the 6.0.5 packages, including platform-server.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package.json
```diff
@@ -12,28 +12,30 @@
...
+    "@angular/platform-server": "6.1.10",
+    "@angular/router": "6.1.10",
     "@nguniversal/module-map-ngfactory-loader": "6.0.0",
     "core-js": "^2.5.4",
     "rxjs": "^6.0.0",
     "zone.js": "^0.8.26",
     "aspnet-prerendering": "^3.0.1",
-    "bootstrap": "^4.1.3"
+    "bootstrap": "^4.1.3",
+    "jquery": "3.3.1",
```
Eilon: Is this a new dependency or is it here to force it back to a specific version?

ryanbrandenburg: jquery and popper.js are both Peer dependencies requestetd by bootstrap that we weren't previously fulfilling.

```
npm WARN bootstrap@4.1.3 requires a peer of jquery@1.9.1 - 3 but none is installed. You must install peer dependencies yourself.
npm WARN bootstrap@4.1.3 requires a peer of popper.js@^1.14.3 but none is installed. You must install peer dependencies yourself.
```

Eilon: OK

Issue #552, opened 2018-06-04
Run NPM tests on CI


ryanbrandenburg:
We should be running tests like [this](https://github.com/aspnet/templating/blob/dev/src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.test.js) on PR's and also on our main builds but currently are not. This allows failures like the one fixed by #539 to go unnoticed.

ryanbrandenburg:
Fixed in #796.

PR #796, opened 2018-10-23
ESLint

### LINKED ISSUE #500: Text based resources shouldn't start with the BOM character to force UTF-8 encoding
default ASP.NET Core Web Application using Kestrel

-`<link rel="stylesheet" href="/css/site.css">`
-`<script src="/js/site.js?v=ji3-IxbEzYWjzzLCGkF1KDjrT2jLbbrSYXw-AhMPNIA"></script>`

[Warn if the byte-order mark](https://sonarwhal.com/docs/user-guide/rules/rule-no-bom/)

ryanbrandenburg:
Fixes #500, #769, #552.


src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package.json
```diff
@@ -8,7 +8,8 @@
     "build:ssr": "ng run Company.WebApplication1:server:dev",
     "test": "ng test",
     "lint": "ng lint",
-    "e2e": "ng e2e"
+    "e2e": "ng e2e",
+    "check": "npm run lint"
```
SteveSandersonMS: I'm not certain why we're adding this `check` script instead of directly running `npm run lint` elsewhere. This is not an objection - I'm fine with it, just saying I'm not sure what the design intent is.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/src/karma.conf.js
```diff
@@ -26,6 +26,6 @@ module.exports = function (config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: ['Chrome'],
-    singleRun: false
+    singleRun: true
```
SteveSandersonMS: I'd think the config here should be the same as whatever is the default in Angular CLI, since this is what ends up in user projects. Is the original default `true`?

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/package.json
```diff
@@ -12,10 +13,26 @@
...
+  },
   "scripts": {
     "start": "rimraf ./build && react-scripts start",
     "build": "react-scripts build",
-    "test": "react-scripts test --env=jsdom",
-    "eject": "react-scripts eject"
+    "test": "cross-env CI=true react-scripts test --env=jsdom",
+    "eject": "react-scripts eject",
+    "lint": "eslint ./src/",
+    "check": "npm run lint&& npm run test"
```
SteveSandersonMS: Minor nitpick: missing the usual space before `&&`

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.js
```diff
@@ -18,3 +16,4 @@ export default class App extends Component {
     );
   }
 }
+App.displayName = App.name;
```
SteveSandersonMS: Is this equivalent to leaving line 9 there but changing it to `static displayName = App.name`?

I don't really know if either way of expressing this is preferable over the other. Perhaps it makes no difference. I'm certainly happy for you to decide what's best.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/Home.js
```diff
@@ -13,7 +11,7 @@ export class Home extends Component {
           <li><a href='https://facebook.github.io/react/'>React</a> for client-side code</li>
           <li><a href='http://getbootstrap.com/'>Bootstrap</a> for layout and styling</li>
         </ul>
-        <p>To help you get started, we've also set up:</p>
+        <p>To help you get started, we&apos;ve also set up:</p>
```
SteveSandersonMS: This is a bit awkward. Maybe we should phrase it as `we have also set up` to avoid the weird apostrophe encoding, so it's easier to read in the source code.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.js
ryanbrandenburg: I like the static thing better, will send an update.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/Home.js
ryanbrandenburg: That's also fine.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/src/karma.conf.js
ryanbrandenburg: The original default is false. I changed it because `singleRun: false` runs in an interactive mode which 1) I find super annoying and 2) Prevents any script from continuing onward. You're correct that we should try to keep it close to the original, so I'll change it back, but I'm sad about it.

src/Microsoft.DotNet.Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.js
```diff
@@ -6,6 +6,7 @@ import { FetchData } from './components/FetchData';
 import { Counter } from './components/Counter';
 
 export default class App extends Component {
+  static displayName = App.name
```
SteveSandersonMS: Ultra-minor nitpick: You have a blank line after the equivalent to this line in other files.

SteveSandersonMS:
Updates look good!

PR #797, opened 2018-10-24
Make AspNetCoreHostingModel element not be insert when targeting net461

jkotalik:
@ryanbrandenburg I assume the TargetFrameworkOverride is only used when targeting net461, correct?


ryanbrandenburg:
Technically when targeting any net framework TFM.

jkotalik:
> net framework TFM

Even better 😄 

@shirhatti I also went ahead and changed inprocess => InProcess, that fine?

shirhatti:
👍

shirhatti:
On second thought, can we verify that the WebSDK isn't case sensitive the other way. Don't want to break something else 😬

jkotalik:
> On second thought, can we verify that the WebSDK isn't case sensitive the other way. Don't want to break something else 😬

Where is this value checked?

jkotalik:
https://github.com/aspnet/websdk/blob/696d0626b5fef5e385bf510a01ee5a7668c4a318/test/Publish/Microsoft.NET.Sdk.Publish.Tasks.Tests/WebConfigTransformTests.cs#L83-L84 Looks like it is tested.

PR #810, opened 2018-10-29
[Design]Fail on Browser warnings

ryanbrandenburg:
For now I just want to see what this does x-plat.


ryanbrandenburg:
We'll take puppeteer before this, so closing.

PR #828, opened 2018-11-09
Use node 10

### LINKED ISSUE #819: Use Node 10
We seem to have updated out of the package which caused problems with node 10, we should upgrade our CI pipelines and korebuild.json

ryanbrandenburg:
Fixes #819.


ryanbrandenburg:
Moving to aspnet/AspNetCore.

PR #834, opened 2018-11-13
Don't tell people to run selenium themselves

ryanbrandenburg:
We don't want to recommend this anymore because it doesn't work.


javiercn:
@ryanbrandenburg Care to elaborate?

ryanbrandenburg:
We try to launch the process [in test fixtures](https://github.com/aspnet/Templating/blob/f82461b4b09898a5eb692e47d99700db7d20ef0e/test/Templates.Test/Infrastructure/SeleniumServerFixture.cs#L71), if it's already been launched it will fail when trying to launch it again.

